PR #1, opened 2020-05-01
Add alias to allow running tests in OSP13

elfiesmelfie:
OSP13 uses docker instead of podman, this change adds an alias for root user so `podman` maps to `docker`, so CI can run against OSP13 and OSP16 without separate branches needed to resolve the differences.

### DIFF
--- a/stf_functional_tests.yml
+++ b/stf_functional_tests.yml
@@ -1,5 +1,16 @@
 #!usr/bin/env ansible-playbook
 ---
+- name: "Add alias for OSP13 to use docker instead of podman"
+  hosts: overcloud_nodes
+  become: true
+  tags:
+    - OSP13
+  tasks:
+    - lineinfile:
+        path: /root/.bashrc
+        line: "alias podman='docker'"
+        create: yes
+
 - name: "Check that collectd is running and generating any metrics"
   hosts: overcloud_nodes
   become: true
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
Testing showed that the alias doesn't work. @lnatapov indicated that it was due to sh being used instead of bash.
New solution is to add an symlink in OSP13 so that podman points to docker. This avoids the need to rewrite the tests and maintain a separate branch for OSP13.

Tested using OSP16, but with the symlink being /usr/bin/podman2->/usr/bin/docker, since podman already existed. The commands in were temporarily modified to use podman2, and some debug statements were included so that the stdout and stderr could be inspected.

Update to test:
```
diff --git a/stf_functional_tests.yml b/stf_functional_tests.yml
index 0fbbb7e..0d82baf 100644
--- a/stf_functional_tests.yml
+++ b/stf_functional_tests.yml
@@ -8,10 +8,17 @@
   tasks:
     - file:
         src: /usr/bin/docker
-        dest: /usr/bin/podman
+        dest: /usr/bin/podman2
         owner: root
         group: wheel
         state: link
+    - name: "test"
+      shell: |
+         podman2 ps
+      register: output
+
+    - debug:
+        var: output.stderr
```

Output when playbook is run:
```
ok: [controller-0] => {
    "output.stderr": "Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg."
}
ok: [compute-0] => {
    "output.stderr": "Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg."
}
```

lnatapov:
looks good

PR #3, opened 2020-05-07
Add summary log file for tests

elfiesmelfie:
Update to @lnatapov's  change to log test summary to file.

This approach uses an ansible callback and produces the same output format.
The file is currently placed in the directory which the playbook is called from.

The callback can be used either by passing a config variable::
ANSIBLE_CALLBACK_WHITELIST=custom_logger ansible-playbook -i `infrared workspace inventory` stf_functional_tests.yml
or by including ``callback_whitelist = custom_logger`` in the [defaults] section of the ansible.cfg

### DIFF
--- a/README
+++ b/README
@@ -32,9 +32,15 @@ OR
 
     ansible-playbook -i `infrared workspace inventory` stf_functional_tests.yml --tags "<one of the tags from above>"
 
+To get a summarised output of the tests at the end, whitelist the provided logging callback.::
+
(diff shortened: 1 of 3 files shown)


elfiesmelfie:
@lnatapov There's one difference between your change and mine, in terms of output: yours places the log file on the undercloud, and mine leaves the file in the local directory. 
This difference can be reconciled either by (a) updating the ``callback_plugins/custom_logger.py`` or (b) addding a line in the jenkins job to copy the logfile to the undercloud.
I prefer option (b) because it's more user friendly for someone running these tests locally during their development or debugging process.

PR #5, opened 2020-08-21
Reorg tests and add config to all reusing of the tests

elfiesmelfie:
Split the tests into separate task files so that they can be imported by other
testing tools.
Currently, it is planned that the collectd_config[1] and tripleo_collectd[2]
roles will use this in molecule and CI testing.
Later on, it'll be used to test other components intended for use with tripleo.

Allowing the changing of container names make it more re-usable as well.

There's a bit of a hack in stf_functional_tests.yaml to let the "default" vars be passed to  the tests; this is intended to be temporary and will be fixed later, it is in place at the moment to keep the behaviour of the stf_functional_tests.yaml identical to before.

[1] https://github.com/infrawatch/collectd-config-ansible-role
[2] https://github.com/infrawatch/tripleo-collectd-ansible-role

### DIFF
--- a/README
+++ b/README
@@ -84,3 +84,14 @@ The task can then be imported in ``stf_functional_tests.yml`` like so::
 ```
 
 
+Configuration
+-------------
+The following vars can be passed to change the bahaviour.
+
+* collectd_container_name
+  The name of the container where collectd is running, e.g. ``collectd-test``
+  default: ``collectd``
+
+* qdr_container_name
+  The name of the container where qdr is running, e.g. ``metrics_qdr``, ``qdr-test``
+  default: ``metrics_qdr``
--- a/README
+++ b/README
@@ -41,18 +41,37 @@ The callback will only report on the status of tasks that have a name beinging w
 
 Note::
     If you haven't deployed using infrared, you can still run the tests if you create your own inventory file, containing one group of hosts called ``overcloud_nodes``.
-
+    Alternatively, you can create your own playbook that imports the tasks in ``tasks/*.yml``
 
 Adding new tests
 ----------------
 
-Adding a new test for existing deployment templates requires no changes in Jenkins.
-Adding tests for new deployment templates requires that the template is in Infrared and deployed in Jenkins.
+Adding a new test for existing deployment templates requires no changes in
+Jenkins.
+Adding tests for new deployment templates requires that the template is in
+Infrared and deployed in Jenkins.
 
-The changes required in this repo for tests is a new role in ``run_stf_tests.yml`` with a tag, e.g.::
+The changes required in this repo for tests is to add additional tasks under
+``tasks/test_<your_test_name>.yml``.
 
 ```
-    - name: Collectd checks
+     - name: "[Test] My test name"
+       shell: |
+           my_test_command
+       register: command_output
+       failed_when: command_output.stdout == some_value
+
+```
+
+Most of the functional tests are a series of shell commands that one would run
+when verifying that particular features were deployed correctly, and can be
+adapted from any manual testing that is done.
+
+
+The task can then be imported in ``stf_functional_tests.yml`` like so::
+
+```
+     - name: Collectd checks
       hosts: overcloud_nodes
       become: true
       tags:
@@ -60,13 +79,8 @@ The changes required in this repo for tests is a new role in ``run_stf_tests.yml
         - collectd-write-qdr-mesh
         - some-other-template-name
       tasks:
-        - name: "[Test] My test name"
-          shell: |
-              my_test_command
-          register: command_output
-          failed_when: command_output.stdout == some_value
+        - import_tasks: tasks/test_<your_test_name>
 
 ```
 
-Most of the functional tests are a series of shell commands that one would run when verifying that particular features were deployed correctly, and can be adapted from any manual testing that is done.
 
--- a/stf_functional_tests.yml
+++ b/stf_functional_tests.yml
@@ -22,35 +22,7 @@
     - collectd-write-qdr-mesh
     - enable-stf
   tasks:
-      - name: "[Test] Check that collectd container is running."
-        shell: |
-          podman ps | grep collectd
-        register: podman_nodes
-        failed_when: podman_nodes.stdout_lines|length != 1
-
-      - name: "[Test] Check for a non-zero number of metrics from collectd"
-        shell: |
-          podman exec collectd collectdctl -s /var/run/collectd-socket listval | wc -l
-        register: num_metrics
-        failed_when: "not (num_metrics.stdout|int !=0 )"
-
-
-      - name: "[Debug] Check for a non-zero number of metrics from collectd"
-        debug:
-           var: num_metrics
-
-      - name: "[Setup] Get the value of some metric from collectd"
-        shell: podman exec collectd collectdctl -s /var/run/collectd-socket listval | tail -1
-        register: met
-
-      - name: "[Debug] Get the value of some metric from collectd"
-        debug:
-          var: met
-
-      - name: "[Test] Get the value of some metric from collectd"
-        shell: podman exec collectd collectdctl -s /var/run/collectd-socket getval {{ met.stdout }}
-        register: stat
-        failed_when: not(stat.stdout_lines|length == 1)
+    - import_tasks: tasks/test_collectd.yml
 
 - name: "Check that metrics_qdr container is running and receiving ANY messages."
   hosts: overcloud_nodes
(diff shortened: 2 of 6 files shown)


README
```diff
@@ -41,32 +41,57 @@ The callback will only report on the status of tasks that have a name beinging w
 
 Note::
     If you haven't deployed using infrared, you can still run the tests if you create your own inventory file, containing one group of hosts called ``overcloud_nodes``.
-
+    Alternatively, you can create your own playbook that imports the tasks in ``tasks/*.yml``
```
leifmadsen, line 44: This line is fine, but the filename should probably be converted to README.md (or README.rst or README.adoc depending on the syntax in use)

README
```diff
@@ -41,32 +41,57 @@ The callback will only report on the status of tasks that have a name beinging w
...
+Jenkins.
+Adding tests for new deployment templates requires that the template is in
+Infrared and deployed in Jenkins.
+
+The changes required in this repo for tests is to add additional tasks under
+``tasks/test_<your_test_name>.yml``.
+
+```
+     - name: "[Test] My test name"
+       shell: |
```
leifmadsen, line 59: Just FYI you might want to suggest using `command` instead of `shell` when `bash` specific commands aren't required as that's the preferred (safer) way to execute things. I wouldn't hold this up though since the risk is pretty low I think.

README
```diff
@@ -41,32 +41,57 @@ The callback will only report on the status of tasks that have a name beinging w
...
+when verifying that particular features were deployed correctly, and can be
+adapted from any manual testing that is done.
+
 
-The changes required in this repo for tests is a new role in ``run_stf_tests.yml`` with a tag, e.g.::
+The task can then be imported in ``stf_functional_tests.yml`` like so::
 
 ```
-    - name: Collectd checks
+     - name: Collectd checks
```
leifmadsen, line 74: This offset I think looks weird? Extra space?

README
```diff
@@ -41,32 +41,57 @@ The callback will only report on the status of tasks that have a name beinging w
...
-          failed_when: command_output.stdout == some_value
+        - import_tasks: tasks/test_<your_test_name>
 
 ```
 
-Most of the functional tests are a series of shell commands that one would run when verifying that particular features were deployed correctly, and can be adapted from any manual testing that is done.
 
+Configuration
+-------------
+The following vars can be passed to change the bahaviour.
```
leifmadsen, line 89: behaviour with a `u` <3

stf_functional_tests.yml
```diff
@@ -13,6 +13,28 @@
         group: wheel
         state: link
 
+- name: "Set up defaults"
```
leifmadsen, line 16: The double quotes are fine, but you don't need them on `name` lines just fyi.

stf_functional_tests.yml
```diff
@@ -13,6 +13,28 @@
...
+  hosts: overcloud_nodes
+  tags: always
+  tasks:
+    - include_vars:
+        file: defaults/main.yml
+        name: play_vars
+
+    - set_fact: "{{item.key}}={{item.value}}"
```
leifmadsen, line 24: Ha! I just had to learn this trick recently!

stf_functional_tests.yml
```diff
@@ -13,6 +13,28 @@
         group: wheel
         state: link
 
+- name: "Set up defaults"
+  hosts: overcloud_nodes
+  tags: always
+  tasks:
+    - include_vars:
+        file: defaults/main.yml
```
leifmadsen, line 21: Do you need to include the default vars? Aren't they available by default?

tasks/test_collectd.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: "[Test] Check that collectd container is running."
+  shell: |
+    podman ps | grep {{ collectd_container_name }}
```
leifmadsen, line 4: Just for portability (it might not be required here?) I'd set a `__container_bin` or something variable and default it to `podman` to allow for override to `docker` if it were ever necessary.

leifmadsen: Although I guess your tests are specifically looking for `podman` output and might not be portable. I'm ok leaving as-is but thought I'd mention it in case you thought it made sense.

stf_functional_tests.yml
elfiesmelfie: Somehow, no, and I'm not certain why, but this was my workaround.

tasks/test_collectd.yml
elfiesmelfie: This is used for both OSP 13 and OSP16 testing, there's a task to create a symlink so that podman -> docker if OSP 13 is used. This might be a better option.

leifmadsen: Ah right I do remember reading that option. Is there a way to check if you're running against an OSP13 vs 16 system? Then you could even dynamically set the appropriate container_bin.

Also you could just add a check in Ansible that looks for availability of the various bins? You'd do that via the `stat` module, then register that results and do a `set_fact` `when: result.stat.exists` or similar.

elfiesmelfie: @Leif, That should go on a separate change IMHO.

csibbitt: > This is used for both OSP 13 and OSP16 testing, there's a task to create a symlink so that podman -> docker if OSP 13 is used. This might be a better option.

We are being asked to remove that symlink, I think. Selecting the correct binary is probably better than symlinking.

elfiesmelfie: @csibbitt that change is in progress, I 'll make the appropriate changes to the docs here, as requests, and then open a separate PR for removing the symlink and selecting the appropriate container_bin instead

README
elfiesmelfie: Opened an issue so I can circle back to this later

tasks/test_collectd.yml
leifmadsen: Yea separate change makes sense to me, just noting it, and it looks like (as Chris noted) to stop doing that :+1: 

Sounds like we're all on the same page.

README
elfiesmelfie: Should be taken care of, I think it was a yank and paste issue

README
leifmadsen: Haha just noticed though it's "bahaviour" so you need to s/ba/be/

stf_functional_tests.yml
leifmadsen: Hrmmm might be worth an issue to investigate why that is, because I'm a bit concerned we may run into some oddities that are hard to debug if this is necessary. Or at least understand why it's necessary and correct.

tasks/test_collectd.yml
elfiesmelfie: Opened an issue, and will address today

README
elfiesmelfie: Issue https://github.com/infrawatch/functional-tests/issues/7  is a reminder to check this

tasks/test_qdr.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: "[Test] Check that qdr container is running."
+  shell: |
+    podman ps | grep {{ qdr_container_name }}
```
EmilienM, line 4: if the container is managed by tripleo, you should use systemdctl to check if the systemd is active and running (which manages the container). Doing grep will certainly leads to failures later; if a container has similar name and such...

elfiesmelfie: That makes sense,  but we still need to check if the container is running when we're using docker on OSP13.

I can add in a check here, and select the proper/systemctl check when podman is used, and use this grep when OSP13/docker are used.

This is part of issue https://github.com/infrawatch/functional-tests/issues/7, when there's a container_bin var to check and use for selection

stf_functional_tests.yml
elfiesmelfie: I opened issue https://github.com/infrawatch/functional-tests/issues/8, I think it's dues to the layout of the directory, and ansible not recognising it as a single role

elfiesmelfie: I fixed that, little mix up on my part with plays vs roles, and how this was originally set up. This is what happens when you write your first ansible scripts and they actually get used.

stf_functional_tests.yml
```diff
@@ -17,73 +16,26 @@
   hosts: overcloud_nodes
   ignore_errors: true
   become: true
+  roles:
```
leifmadsen, line 19: I feel like use of `roles` is out of favour with a preference to doing an `import_role` or `include_role` as a task.

In which case, I think you could do a lookup of the files and iterate over the list with `loop` and `include_role: "{{ item }}"`. 

Not a big deal, more a style thing.

elfiesmelfie: The role is imported here to do set-up, so that the defaults are available to the tasks being run.

There's not anything to loop through really.

leifmadsen: Well I was just saying you can loop through the roles by importing them via `import_role` vs the `roles` parameter which seems to be falling out of favour.

Ultimately this is a style decision though, and I have no strong preference here. If it works as-is, good enough :)

PR #10, opened 2020-08-31
Update stf_functional_tests.yml

elfiesmelfie:
When other tags are passed, the set up tasks are not run, these should have the "always" tag to ensure they run.


lnatapov (approved):
looks good

PR #11, opened 2020-09-18
Update formatting on README

elfiesmelfie:


leifmadsen:
I think you need a newline after `Note::`

leifmadsen:
Also is `OR::` the intended content?

README.rst
```diff
@@ -28,16 +28,16 @@ If you have deployed client-side STF and want to test it locally, you can run th
...
+To get a summarised output of the tests at the end, use the provided logging callback::
 
     ANSIBLE_CALLBACK_WHITELIST=custom_logger ansible-playbook -i `infrared workspace inventory` stf_functional_tests.yml
 
-The logging callback will summarise the tests run on each node and whether they passes or failed.
-The callback will only report on the status of tasks that have a name beinging with "[Test]".
+The logging callback will summarise the tests run on each node and whether they passed or failed.
+The callback will only report on the status of tasks that have a name beginning with ``[Test]``.
 
 Note::
```
leifmadsen, line 42:
```suggestion
Note::

```

README.rst
```diff
@@ -28,16 +28,16 @@ If you have deployed client-side STF and want to test it locally, you can run th
 
     ansible-playbook -i `infrared workspace inventory` stf_functional_tests.yml
 
-OR
+OR::
```
leifmadsen, line 31:
```suggestion
OR:
```

README.rst
elfiesmelfie: Actually, it should be ``.. note::`` without a newline. The ``::`` is a mistake originally

README.rst
elfiesmelfie: I want the line that follows to be formatted as a literal. Only one of the colons gets into the rendered text

leifmadsen (approved):

README.rst
```diff
@@ -28,18 +28,18 @@ If you have deployed client-side STF and want to test it locally, you can run th
...
 
     ANSIBLE_CALLBACK_WHITELIST=custom_logger ansible-playbook -i `infrared workspace inventory` stf_functional_tests.yml
 
-The logging callback will summarise the tests run on each node and whether they passes or failed.
-The callback will only report on the status of tasks that have a name beinging with "[Test]".
+The logging callback will summarise the tests run on each node and whether they passed or failed.
+The callback will only report on the status of tasks that have a name beginning with ``[Test]``.
 
-Note::
+.. note::
```
csibbitt, line 42: This still doesn't look like it renders how you want

csibbitt (approved):
Read the rendered version and it seemed good enough for me

README.rst
elfiesmelfie: I think we need some sphinx build of style sheets to do this properly, but this will have it indicated as a note. It's good enough for the moment.

README.rst
elfiesmelfie: I am no longer familiar with it, which is why I had to do this patch -- to fix what I thought I knew

leifmadsen: That's why I always just end up with markdown, I always forget all the weird syntax rules of RST.

README.rst
leifmadsen: Also looking at the examples I have found, the note should start on the same line as the admonition is defined:

.. note:: If you haven't deployed using infrared, you can still run the tests if you create your own inventory file, containing one group of hosts called ``overcloud_nodes``.
   Alternatively, you can create your own playbook that imports the tasks in ``tasks/*.yml``

leifmadsen: And of course I found _another_ example. All the indenting looks like it should align to 3 spaces in (alignment on the newline starting at the `n` of note.

Anyways, more just a curiosity, go ahead and merge when you see fit :)

leifmadsen (approved):

PR #16, opened 2020-12-09
Adding networks and routers,quota,users tasks to the OSP post install…

lnatapov:
I have added more OSP tasks to the  openstack post_install

### DIFF
--- a/roles/post_install/openstack_tasks/tasks/main.yml
+++ b/roles/post_install/openstack_tasks/tasks/main.yml
@@ -30,6 +30,20 @@
   import_tasks: instances.yml
   when: instance and resource_state == 'present'
 
+- name: Networks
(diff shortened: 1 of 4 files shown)


roles/post_install/openstack_tasks/tasks/main.yml
```diff
@@ -30,6 +30,20 @@
   import_tasks: instances.yml
   when: instance and resource_state == 'present'
 
+- name: Networks
+  tags: network
+  import_tasks: networks
```
elfiesmelfie, line 35: Should this be `networks_and_routers.yml`?

lnatapov: Thanks ! Fixed

PR #21, opened 2021-01-05
WIP: End2End STF functional tests

lnatapov:
End2End STF functional tests. 
Querying STF server for collectd and ceilometer metrics and events

### DIFF
--- a/openstack_tasks.yml
+++ b/openstack_tasks.yml
@@ -5,3 +5,12 @@
   gather_facts: False
   roles:
     - post_install/openstack_tasks
+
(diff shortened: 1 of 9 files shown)


leifmadsen:
Note that these scripts are expected to be run from a smoketest container (job) on the OpenShift instance running STF, and not from a remote session.

lnatapov:
> Note that these scripts are expected to be run from a smoketest container (job) on the OpenShift instance running STF, and not from a remote session.

Thanks ! Yes,I know. I will use them as example to create remote queries

PR #22, opened 2021-01-12
Add -C to grep commands for busybox compat

csibbitt:
These tests were failing in our GHA-based CI because it has busybox instead of GNU grep
https://github.com/infrawatch/tripleo-qdr-ansible-role/runs/1689224720#step:5:404

### DIFF
--- a/tasks/test_qdr.yml
+++ b/tasks/test_qdr.yml
@@ -7,7 +7,7 @@
 
 - name: "[Setup] Get Qdr bus address"
   shell: |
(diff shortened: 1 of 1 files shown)


csibbitt:
Just FYI for you two. I'm testing this now and will merge it if it fixes my problem

csibbitt:
Works, see https://github.com/infrawatch/tripleo-qdr-ansible-role/runs/1689570976#step:5:319

PR #24, opened 2021-04-25
Changes to E2E testing

lnatapov:


paramite (approved):
Seems legit.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+  changed_when: no
+
+- name: "[Debug] Check whether collectd cpu ceph metrics received in prometheus for all ceph nodes"
+  debug:
+    var: checkmyconf
+
+
+- name: "[Test] Check whether collectd cpu ceph metrics log created and received  in Prometheus"
+  debug: msg="COLLECTD CPU CEPH METRICS RECEIVED FOR ALL CEPH NODES"
+  when: checkmyconf.rc == 0
```
elfiesmelfie, line 47: You can replace these three tasks with a single task that used the failed_when option for the task.
```
- name: "[Test] Check whether collectd cpu ceph metrics received in prometheus for all ceph nodes"
  command: egrep ceph-0 /tmp/cpu_ceph_bytes
  register: checkmyconf
  check_mode: no
  failed_when:
    - check_my_conf.rc != 0
```

IMHO, this update is easier to grok, and makes it easier to see the different tests, and does what you want, as far as I can tell.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+  check_mode: no
+  changed_when: no
+
+- name: "[Debug] Check whether collectd interface TX total  metrics received  in prometheus for all overcloud nodes"
+  debug:
+    var: checkmyconf
+
+- name: Check whether collectd interface TX total metrics log created and received in Prometheus.
+  debug: msg="COLLECTD INTERFACE TOTAL TX  METRICS RECEIVED FOR ALL OVERCLOUD NODES"
+  when: checkmyconf.rc == 0
```
elfiesmelfie, line 61: These tasks can be combined into a single task using ``failed_when``:

```
- name: "[Test] Check whether collectd interface TX total  metrics received  in prometheus for all overcloud nodes"
  command: egrep 'controller-0|controller-1|controller-2|compute-0|compute-1|ceph-0' /tmp/collectd_interface_tx_total
  register: checkmyconf
  check_mode: no
  changed_when: no
  failed_when:
    - checkmyconf.rc != 0
```

If you want to keep the debug,

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+    src: "{{ item }}"
+    dest: "/tmp"
+    owner: root
+    group: root
+    mode: 0755
+  with_items:
+    ['collectd_prometheus_query.sh']
+
+- name: execute promethues collectd query 
+  shell: "/bin/bash /tmp/collectd_prometheus_query.sh &"
```
elfiesmelfie, line 12: Does this have to be run in this manner? Or can the process be run in the foreground so that the STDOUT can be captured and used for the tests? It might be useful to know that this task ran successfully, for example, in case a failure in this would result in failures in subsequent tasks.

To make sure the command is run successfully (i.e. without errors), you can use the failed_when option for the task 
```
- name: execute promethues collectd query 
  shell: "/bin/bash /tmp/collectd_prometheus_query.sh &"
    register: checkmyconf
    failed_when:
      - checkmyconf.rc != 0
```

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+  changed_when: no
+
+- name: "[Debug] Check whether collectd cpu metrics log created and received in Prometheus."
+  debug:
+    var: checkmyconf
+
+- name: "[Test] Check whether collectd cpu metrics log created and received in Prometheus"
+  debug: msg="COLLECTD CPU METRICS RECEIVED FOR ALL OVERCLOUD NODES"
+  when: checkmyconf.rc == 0
+
```
elfiesmelfie, line 32: These three tasks could be consolidated into a single task that fails when the condition is not met.
```
- name: "[Test] Check whether collectd cpu metrics received in prometheus for all overcloud nodes"
   command: egrep 'controller-0|controller-1|controller-2|compute-0|compute-1|ceph-0' /tmp/query_collectd_cpu_percent
  register: checkmyconf
  check_mode: no
  changed_when: no
  failed_when:
    - checkmyconf.rc != 0
```

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
+- name: copy shell script that executes prometheus query for collectd metrics
+  copy:
+    src: "{{ item }}"
+    dest: "/tmp"
+    owner: root
+    group: root
+    mode: 0755
+  with_items:
+    ['collectd_prometheus_query.sh']
```
elfiesmelfie, line 9: Suggestion: instead of creating this script, copying it over, running it and then examining the files, could you execute each of the commands directly using the command module?
That way, there's less chance of the play failing due to unrelated (i.e. copy, running or writing the files) reasons. In running the script, any of the curl commands could fail, and you wouldn't know which or why.
If the curl commands are run directly, and one set of metrics was unavailable, then the others will still be run instead of seeing that the `collectd_prometheus_query.sh` script failed.

ansible.cfg
```diff
@@ -0,0 +1,23 @@
...
+
+retry_files_enabled = False
+
+[paramiko_connection]
+record_host_keys = False
+
+[ssh_connection]
+pipelining = True
+ssh_args = -o ControlMaster=auto -o ControlPersist=60s
+control_path = %(directory)s/%%h-%%r
```
elfiesmelfie, line 23: These should be set in your local environment, as opposed to here. 
IMHO, this file should be removed

elfiesmelfie (commented):
Please remove the callback_plugins/custom_logger.pyc file.

Following the conventions of the other tests in this repo, you can use the ``ignore_errors: true`` option for the play in ``stf_functional_tests.yml``, so that the failed tests will not cause the play to end. When you do this, you can use ``failed_when`` in tasks and the custom_logger will summarise the results for you. As it stands at the moment, the play will succeed regardless of whether the tests pass or fail, since all the tasks will be passed.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+  register: checkmyconf
+
+- name: "[Debug] Execute Prometheus collectd query"
+  debug:
+    var: checkmyconf
+
+- name: "[Test] Check whether collectd cpu metrics received in prometheus for all overcloud nodes"
+  command: egrep 'controller-0|controller-1|controller-2|compute-0|compute-1|ceph-0' /tmp/query_collectd_cpu_percent
+  register: checkmyconf
+  check_mode: no
```
csibbitt, line 22: Why do you set check_mode no here?

tasks/test_e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+  changed_when: no
+
+- name: "[Debug] Check whether collectd cpu metrics log created and received in Prometheus."
+  debug:
+    var: checkmyconf
+
+- name: "[Test] Check whether collectd cpu metrics log created and received in Prometheus"
+  debug: msg="COLLECTD CPU METRICS RECEIVED FOR ALL OVERCLOUD NODES"
+  when: checkmyconf.rc == 0
```
csibbitt, line 31: It's not clear what task is supposed to fail when checkmyconf.rc != 0. I like Emma's suggestions.

csibbitt (changes requested):
My only major concern is I think that "[Test] Check whether collectd * metrics received..." tasks may not actually fail when the egrep fails, as Emma noted.

The "cpu ceph metrics" thing also confused me for a bit, but is ultimately not harmful I guess.

lnatapov:
@csibbitt @elfiesmelfie please,review my latest commit
https://github.com/infrawatch/functional-tests/pull/24/commits/2253dbf32c4d0716e4fca7f6f452a70af4a6bf32

stf_functional_tests.yml
```diff
@@ -21,15 +21,20 @@
...
       tags:
-        - collectd-write-qdr-edge-only
-        - collectd-write-qdr-mesh
-        - ceilometer-write-qdr-edge-only
-        - ceilometer-write-qdr-mesh
         - enable-stf
+        - stf-connectors-osp13
+
+- name: "Run e2e tests"
+  hosts: undercloud-0
```
csibbitt, line 34: I think you should add `ignore_errors: true` here, otherwise the tests could end prematurely when one of the e2e tests fails. I think this is what Emma describes in her top-level review comment.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+      shell: grep interconnect /home/stack/virt/stf-connectors-osp16-edge.yaml | sed 's;.*\(-service-telemetry.*\);prometheus-operated\1;'
+      register: prom_url
+      failed_when: "prom_url.stdout_lines|length == 0"
+    - debug:
+        var: prom_url.stdout_lines
+    
+    - debug: 
+        msg: The URL is {{ prom_url.stdout_lines }}
+
+    - name: "Capture ceph_cpu_bytes metrics"
```
csibbitt, line 17: ceph_ceph_bytes

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+      shell: /usr/bin/curl -k http://{{ prom_url.stdout_lines[0] }}/api/v1/query? --data-urlencode 'query=collectd_ceph_ceph_bytes {plugin_instance="ceph-osd.0"}[1m]' > /tmp/query_ceph_cpu_bytes
```
csibbitt, line 18: ceph_ceph_bytes

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+        - checkmyconf.rc !=0
+    
+    - name: "Capture collectd_iinterface_tx_total"
+      shell: /usr/bin/curl -k http://{{ prom_url.stdout_lines[0] }}/api/v1/query? --data-urlencode 'query=collectd_interface_if_packets_tx_total {type_instance="base"}[1m]' > /tmp/query_collectd_interface_tx_total
+      register: checkmyconf
+      failed_when:
+        - checkmyconf.rc !=0
+
+    - name: "[Test] Check whether ceph cpu  metrics received in prometheus for all ceph nodes"
+      command: egrep 'ceph' /tmp/query_ceph_cpu_bytes
```
csibbitt, line 36: ceph_ceph_bytes

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+      failed_when:
+        - checkmyconf.rc !=0
+    
+    - name: "Capture collectd_iinterface_tx_total"
+      shell: /usr/bin/curl -k http://{{ prom_url.stdout_lines[0] }}/api/v1/query? --data-urlencode 'query=collectd_interface_if_packets_tx_total {type_instance="base"}[1m]' > /tmp/query_collectd_interface_tx_total
+      register: checkmyconf
+      failed_when:
+        - checkmyconf.rc !=0
+
+    - name: "[Test] Check whether ceph cpu  metrics received in prometheus for all ceph nodes"
```
csibbitt, line 35: This says "**all** ceph nodes" but the grep looks like it would be satisfied by metrics from "**any** ceph node". Checking from All would be preferable - is that really what's happening here? I haven't looked at the output in a live system - could we count the number of lines in that file and expect it to match the number of ceph nodes?

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+      command: egrep 'ceph' /tmp/query_ceph_cpu_bytes
+      register: checkmyconf
+      changed_when: no
+      failed_when:
+        - checkmyconf.rc != 0
+    - debug:
+        msg="COLLECTD CPU CEPH METRICS RECEIVED FOR ALL CEPH NODES"
+
+    - name: "[Test] Check whether collectd interface TX total  metrics received  in prometheus for all overcloud nodes"
```
csibbitt, line 44: Same all vs. any problem here, I think.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+        msg="COLLECTD INTERFACE TOTAL TX  METRICS RECEIVED FOR ALL OVERCLOUD NODES"
+
+    - name: "[Test] Check whether collectd cpu metrics log created and received in Prometheus"
+      command: egrep 'controller-0|controller-1|controller-2|compute-0|compute-1|ceph-0' /tmp/query_collectd_cpu_percent
+      register: checkmyconf
+      changed_when: no
+      failed_when:
+        - checkmyconf.rc != 0
+    - debug:
+        msg="COLLECTD CPU METRICS RECEIVED FOR ALL OVERCLOUD NODES"
```
csibbitt, line 60: Same all vs. any problem on this one too

csibbitt (commented):
Thanks for cleaning up the structure of this, it seems more robust and easier to follow now. I think there is still one step missing for correct error tracking (ignore_errors), and I realized upon looking closer at the actual tests that they perhaps pretend to be stricter than they are (all nodes vs any node).

My guess from looking at the queries is that you could confirm there to be a specific # of lines in the output file (or # of occurrences of a substring in the output or something) to check that ALL nodes are actually reporting in. This seems preferable to simply correcting the output to say "any node" :) If I've misunderstood and theses tests really do confirm that every node produces metrics, then maybe a short snippet of the expected output in the comments would clarify how it work.

stf_functional_tests.yml
lnatapov: Thanks,Chris ! Will add it. Here is how output of ceph query looks like:
http://pastebin.test.redhat.com/960822
Basically we need to see if we get some metrics. In case when there is no metrics the query will return empty list and grep will fail.

lnatapov: > I think you should add `ignore_errors: true` here, otherwise the tests could end prematurely when one of the e2e tests fails. I think this is what Emma describes in her top-level review comment.

Done.

tasks/test_e2e.yml
lnatapov: > This says "**all** ceph nodes" but the grep looks like it would be satisfied by metrics from "**any** ceph node". Checking from All would be preferable - is that really what's happening here? I haven't looked at the output in a live system - could we count the number of lines in that file and expect it to match the number of ceph nodes?

Count  lines in the query output won't help since it's not related to a number of ceph nodes.
Yea,you are right grep will satisfy "any". The queries are based on the known number of overcloud nodes.
For example  I know in advance how many overcloud nodes will be sending metrics e.g. 3 controllers,2 computes and 1 ceph.  Probably there is a way to check how many overcloud nodes we have before running the query and dynamically update query according to every time we run it it but I think for now it's enough. Let's have something and than will think how to enhance it.

csibbitt (approved):

tasks/test_e2e.yml
elfiesmelfie: Since checking any is okay for now, I suggest updating the task name to reflect ANY vs all.

tasks/test_e2e.yml
elfiesmelfie: Ditto, update the task name

lnatapov: > Ditto, update the task name

Done

tasks/test_e2e.yml
lnatapov: > Since checking any is okay for now, I suggest updating the task name to reflect ANY vs all.

Done.

tasks/test_e2e.yml
```diff
@@ -0,0 +1,86 @@
...
+        msg="COLLECTD CPU METRICS RECEIVED FOR ANY OVERCLOUD NODES"
+
+
+    - name: "[Test] Check whether collectd memory metrics log created and received in Prometheus"
+      command: egrep 'controller-0|controller-1|controller-2|compute-0|compute-1|ceph-0' /tmp/query_collectd_memory
+      register: checkmyconf
+      changed_when: no
+      failed_when:
+        - checkmyconf.rc != 0
+    - debug:
```
elfiesmelfie, line 75: This debug message will output regardless of whether the previous task failed or succeeded.
I would suggest either adding a `when` to this debug or splitting the previous task and doing the checking in an `[assert](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/assert_module.html)` task, using a success_msg option

e.g.
```
   - name: "[Set-up] Check whether collectd memory metrics log created and received in prometheus"
      ...
      changed_when: no
    - name: "[Test] Check whether collectd memory metrics log created ans received in Prometheus"
      assert:
        that:
            - checkmyconf.rc == 0
        success_msg: "COLLECTD MEMORY METRICS RECEIVED FOR ANY OVERCLOUD NODES"

```

This is applicable for each test/debug task pair.
My concern is that outputting that debug message will cause confusion if it is always printed.

Another option is to remove the debug, but I suspect it was there for a reason.

elfiesmelfie (changes requested):
Can you take a look at the debug messages that are being printed after each test. They are going to cause confusion unless they're tied to the outcome of the tests.

elfiesmelfie (approved):

PR #31, opened 2022-12-20
Removing hardcoded Prometheus username and password

lnatapov:
Currently in order to run Prometheus queries on the OCP cluster we are using hardcoded prometheus username and password that is basically exposed to everyone.

A new code will obtain these credentials by accessing OCP and store it in variables ,so it won't be exposed.

### DIFF
--- a/stf_functional_tests.yml
+++ b/stf_functional_tests.yml
@@ -45,6 +45,13 @@
   tasks:
     - name: "Run collectd Prometheus queries"  
       import_tasks: tasks/test_e2e.yml
+      vars:
+        prom_user: "{{ ( prom_secret.stdout_lines[0] | b64decode | split(':'))[0] }}"
+        prom_pass: "{{ prom_secret.stdout_lines[1] | b64decode }}"
+        prom_url: "{{ prom_route.stdout }}"
+        stf_user: "quicklab"
+        stf_host: "{{ stf_fqdn.stdout }}"
+
       tags:
         - custom-stf-ceph
         - stf-connectors-osp13-ffu
--- a/tasks/test_e2e.yml
+++ b/tasks/test_e2e.yml
@@ -1,39 +1,66 @@
-    - name: "Extract URL from stf-connectors template"
-      shell: grep interconnect /home/stack/virt/stf-connectors* | sed 's;.*\(-service-telemetry.*\);default-prometheus-proxy\1;'
-      register: prom_url
-      failed_when: "prom_url.stdout_lines|length == 0"
+    - name: "Copy quicklab key"
+      shell: curl https://gitlab.cee.redhat.com/cee_ops/quicklab/raw/master/docs/quicklab.key --output ~/.ssh/quicklab.key; chmod 600 ~/.ssh/quicklab.key
+          
+    - name: "Extract STF FQDN"
+      shell: grep interconnect /home/stack/virt/stf-connectors* | sed '/\:*\./s/^[^.]*\ //'
+      register: stf_fqdn
     - debug:
-        var: prom_url.stdout_lines
+        var: stf_fqdn.stdout
+ 
+    - name: Get the default-prometheus-htpasswd secret
+      shell:
+        cmd: |
+          ssh -i ~/.ssh/quicklab.key   {{ stf_user }}@{{ stf_host }}  oc get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
+      register: prom_secret
+
+  
+    - debug:
+        var: prom_secret | string
+
+    - name: "Get the prom URL"
+      shell:
+        cmd: |
+          ssh -i ~/.ssh/quicklab.key  {{ stf_user }}@{{ stf_host }}  oc get route/default-prometheus-proxy -ojson | jq '.spec.host' | sed 's/"//g'
+      register: prom_route
 
     - debug:
-        msg: The URL is {{ prom_url.stdout_lines }}
+        var:  prom_route.stdout
 
+    - name: "Get the prom creds from the secret"
+      set_fact:
+        prom_user: "{{ ( prom_secret.stdout_lines[0] | b64decode | split(':'))[0] }}"
+        prom_pass: "{{ prom_secret.stdout_lines[1] | b64decode }}"
+
+    - debug: var=prom_user
+
+
+ 
     - name: "RHELOSP-37759"
-      shell: /usr/bin/curl -k -u "internal:1tp9zXur-t8Ve,wPkVgE" -g https://{{ prom_url.stdout_lines[0] }}/api/v1/query? --data-urlencode 'query=collectd_cpu_percent {plugin_instance="0"}[1m]' > /tmp/query_collectd_cpu_percent
+      shell: /usr/bin/curl -k -u "{{ prom_user}}:{{ prom_pass }}" -g https://{{ prom_url }}/api/v1/query? --data-urlencode 'query=collectd_cpu_percent {plugin_instance="0"}[1m]' > /tmp/query_collectd_cpu_percent
       register: checkmyconf
       failed_when:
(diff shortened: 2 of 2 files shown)


tasks/test_e2e.yml
```diff
@@ -1,39 +1,66 @@
...
 
+  
     - debug:
-        msg: The URL is {{ prom_url.stdout_lines }}
+        var: prom_secret | string
 
+    - name: "Get the prom URL"
+      shell:
+        cmd: |
+          ssh -i ~/.ssh/quicklab.key  {{ stf_user }}@{{ stf_host }}  oc get route/default-prometheus-proxy -ojson | jq '.spec.host' | sed 's/"//g'
```
elfiesmelfie, line 23: This can be configured as a host in an inventory file. Currently, these are run on the group ``overcloud_nodes``

It makes sense to add a node which can access the ``oc`` too and kubeconfig. Each host can be in multiple groups.

the file will look something like this

```
# ... other groups here
# node with access to the OpenShift cluster
[stf]
<hostname> ansible_host="<ipaddress>" ansible_user="<user>" ansible_ssh_private_key_file="<path/to/private/key"
```

In this case it would look like this:
```
[stf]
my_stf_hostname ansible_host="$stf_host" ansible_user="$stf_user" ansible_ssh_private_key_file="/home/user/.ssh/quicklab.key"
```

stf_functional_tests.yml
```diff
@@ -45,6 +45,13 @@
   tasks:
     - name: "Run collectd Prometheus queries"  
       import_tasks: tasks/test_e2e.yml
+      vars:
+        prom_user: "{{ ( prom_secret.stdout_lines[0] | b64decode | split(':'))[0] }}"
```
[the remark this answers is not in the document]
lnatapov: it's in test_e2e.yml
----
- name: "Copy quicklab key"
      shell: curl https://gitlab.cee.redhat.com/cee_ops/quicklab/raw/master/docs/quicklab.key --output ~/.ssh/quicklab.key; chmod 600 ~/.ssh/quicklab.key

    - name: "Extract STF FQDN"
      shell: grep interconnect /home/stack/virt/stf-connectors* | sed '/\:*\./s/^[^.]*\ //'
      register: stf_fqdn
    - debug:
        var: stf_fqdn.stdout

    - name: Get the default-prometheus-htpasswd secret
      shell:
        cmd: |
          ssh -i ~/.ssh/quicklab.key   {{ stf_user }}@{{ stf_host }}  oc get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
      register: prom_secret


    - debug:
        var: prom_secret | string

    - name: "Get the prom URL"
      shell:
        cmd: |
          ssh -i ~/.ssh/quicklab.key  {{ stf_user }}@{{ stf_host }}  oc get route/default-prometheus-proxy -ojson | jq '.spec.host' | sed 's/"//g'
      register: prom_route

    - debug:
        var:  prom_route.stdout

    - name: "Get the prom creds from the secret"
      set_fact:
        prom_user: "{{ ( prom_secret.stdout_lines[0] | b64decode | split(':'))[0] }}"
        prom_pass: "{{ prom_secret.stdout_lines[1] | b64decode }}"

    - debug: var=prom_user

tasks/test_e2e.yml
lnatapov: e2e tests (prometheus queries) run on undercloud

elfiesmelfie: In these tasks, because you're using ssh in the shell command, this is equivalent to running them on the stf_host.

stf_functional_tests.yml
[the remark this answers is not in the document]
elfiesmelfie: Those vars wouldn't yet exist in this playbook. They are defined in the test_e2e task which is being imported.

tasks/test_e2e.yml
lnatapov: Hey,Emma ! Regarding inventory file. When we run functional tests it uses IR inventory file. How can I add STF host in the inventory ? Also if I understood correctly if we add it to inventory the name of the host will be hardcoded ,so if I will have different  cluster with different name it won't work.

elfiesmelfie: You can add multiple inventory files for ansible, so there wouldn't be a need to update the infrared one.

For changing hostnames, you would need a new inventory file for each deployment of STF, but that would be equivalent to using a different inventory file per infrared deployment as well. 
The test run shouldn't care about it, and just use the inventories that are pointed at it. We would ideally have  pre-described group names i.e. overcloud, undercloud, compute, controller (like in the infrared inv file) and stf (as proposed above, in the stf.inv file.

lnatapov: Thanks,Emma. I submitted  the requested change

README.rst
```diff
@@ -93,4 +93,15 @@ The following vars can be passed to change the behaviour.
...
+  default: ``metrics_qdr`
+
+
+
+
+Running e2e tests against existing STF
+--------------------------------------
+
+E2E tests are currently running  against predeployed STF that installed  on top of OpenShift deployed in QuickLab insfrastracture.
+There is an inventory file called stf.inf that should be uncluded in the ansible-playbbok command along side with the Infrared inventory file.
```
elfiesmelfie, line 105:
```suggestion
There is an inventory file called stf.inf that should be included in the ansible-playbook command along side with the Infrared inventory file.
```

stf.inv
```diff
@@ -0,0 +1,3 @@
+[stf]
```
elfiesmelfie, line 1: This file is not needed, if you include a note that people need to add an inventory file., and list the groups that are needed in that file.

You can reference a default.inv with those groups already defined, but commented hostnames, such as in [1].


[1] https://github.com/opnfv/barometer/blob/master/docker/ansible/default.inv

tasks/test_e2e.yml
```diff
@@ -1,39 +1,80 @@
...
       register: checkmyconf
       failed_when:
         - checkmyconf.rc !=0
 
     - name: "RHELOSP-57528"
-      shell: /usr/bin/curl -k -u "internal:1tp9zXur-t8Ve,wPkVgE" -g https://{{ prom_url.stdout_lines[0] }}/api/v1/query? --data-urlencode 'query=collectd_ceph_ceph_bytes {plugin_instance="ceph-osd.5"}[1m]' > /tmp/query_ceph_ceph_bytes
+      shell:
+        cmd: >-
+          /usr/bin/curl -k -u "{{ prom_user }}:{{ prom_pass }}" \
+            -g https://{{ prom_url }}/api/v1/query? \
```
elfiesmelfie, line 42: Can you update the indentation so these lines are aligned? 

First line is okay, and each subsequent line should be at the same indentation as the second one.
Same comment for the other shell tasks.

e.g. 
```
  /usr/bin/curl -k -u "{{ prom_user }}:{{ prom_pass }}" \
    -g https://{{ prom_url }}/api/v1/query? \
    --data-urlencode 'query=collectd_ceph_ceph_bytes {plugin_instance="ceph-osd.5"}[1m]' \
    --output /tmp/query_ceph_ceph_bytes
```

elfiesmelfie (changes requested):

default.inv
```diff
@@ -0,0 +1,63 @@
...
+[prometheus_hosts]
+#localhost
+
+[zookeeper_hosts]
+#NOTE: currently one zookeeper host is supported
+#hostname
+
+[kafka_hosts]
+#hostname
+
```
elfiesmelfie, line 60:
```suggestion
```

elfiesmelfie (changes requested):
Looks good except for extra sections in the default.inv

default.inv
lnatapov: Removed

elfiesmelfie (approved):

PR #33, opened 2023-01-09
Functional tets adjustments

lnatapov:
1.Adding task for obtaining OCP key file
2.Changing download location of OCP key file
3.Adding "{{ variable.split }}" to work with lower versions of ansible that IR uses.

### DIFF
--- a/stf_functional_tests.yml
+++ b/stf_functional_tests.yml
@@ -36,6 +36,12 @@
         - stf-connectors-osp13-ffu
         - gnocchi-connectors-swift
 
+- name: "Obtain OCP key file"
+  hosts: localhost
+  tasks:
+    - name: "Download OpenShift key file"
+      shell: curl -k https://gitlab.cee.redhat.com/cee_ops/quicklab/raw/master/docs/quicklab.key --output ~/.ssh/quicklab.key; chmod 600 ~/.ssh/quicklab.key
+
 - name: "Run e2e tests"
   hosts: stf
   ignore_errors: true
--- a/default.inv
+++ b/default.inv
@@ -49,5 +49,5 @@
 #ceph-0
 
 [stf]
-quicklab  ansible_host="10.0.91.89" ansible_user="quicklab" ansible_ssh_private_key_file="/home/quicklab/.ssh/quicklab.key"
+quicklab  ansible_host="10.0.91.89" ansible_user="quicklab" ansible_ssh_private_key_file=" ~/.ssh/quicklab.key"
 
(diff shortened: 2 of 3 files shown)


stf_functional_tests.yml
```diff
@@ -36,6 +36,12 @@
         - stf-connectors-osp13-ffu
         - gnocchi-connectors-swift
 
+- name: "Obtain OCP key file"
+  hosts: localhost
+  tasks:
+    - name: "Download OpenShift key file"
+      shell: curl -k https://gitlab.cee.redhat.com/cee_ops/quicklab/raw/master/docs/quicklab.key --output ~/.ssh/quicklab.key; chmod 600 ~/.ssh/quicklab.key
+
```
elfiesmelfie, line 44: We don't want to do this here. This is something that users should be instructed to do before running the playbooks. 

These roles are intended to be re-used in multiple place, downstream as well as upstream testing

lnatapov: We need this in order to be able to run OC commands on the OCP cluster.
Currently downstream tests run against quicklab cluster. The functional tests stage is a part of the jenkins jobs and we can't manually download the key file before running the tests. This file should be present on Jenkins slave which is used to run the playbooks and it's a different machine for every deployment.

lnatapov: Currently functional tests that are outside sif-verify-containers are tailored to run against quicklab cluster. In the future we won't need it but while we are working on the unified job that will include deployment of STF and OSP we have to continue to run tests against permanently deployed quicklab cluster. The goal is to eliminate usage of prometheus username and password that was hardcoded in fucntional tests. The solution is to run queries directly on the OCP quicklab machine but in order to do that we need this file

elfiesmelfie: We shouldn't be putting internal addresses upstream.

I would suggest adding a tag called ``download_ssh_key`` and also passing a var downstream called ``remote_key_uri``.
This can than be changed to  

```
 - name: "Obtain OCP key file"
  hosts: localhost
  when: remote_key_uri is defined
  tasks:
    - name: "Download OpenShift key file"
      shell: curl -k {{ remote_key_uri }} --output ~/.ssh/quicklab.key; chmod 600 ~/.ssh/openshift.key
```

Following this, you would need to add ``-e remote_key_uri=... `` to the invocation command downstream, and this task will only run if you define that var.

alternatively, You can set a default for the value (remote_key_uri: '') and use ``when: len(remote_key_url)>0`` instead. You still need to pass the var with -e.

PR #37, opened 2023-01-28
Adding HA role and tests

lnatapov:
1.Adding HA role and tests
2.Removing duplicate task from snmpTraps tests

### DIFF
--- a/roles/test_ha/tasks/main.yml
+++ b/roles/test_ha/tasks/main.yml
@@ -6,8 +6,10 @@
 - name: "Add highAvailability.enabled: true to the spec section"
   ansible.builtin.shell:
     cmd: |
-      oc patch stf/default --type merge -p '{"spec":{"highAvailability":{"enabled": true}}}' ; sleep 120
+      oc patch stf/default --type merge -p '{"spec":{"highAvailability":{"enabled": true}}}' && sleep 120
   changed_when: false
+
+
 
 - name: "Verify that AMQP Interconnect shows two containers"
   ansible.builtin.shell:
@@ -15,7 +17,9 @@
       oc get pods | grep default-interconnect| wc -l
   register: output
   failed_when: "output.stdout|int !=2"
+  changed_when: false
   
+
 
 - name: "Verify that Alert Manager  shows two containers"
   ansible.builtin.shell:
@@ -23,6 +27,8 @@
       oc get pods | grep alertmanager| wc -l
   register: output
   failed_when: "output.stdout|int !=2"
+  changed_when: false
+
 
 
 - name: "Verify that Prometheus shows two containers"
@@ -31,6 +37,8 @@
       oc get pods | grep prometheus-default| wc -l
   register: output
   failed_when: "output.stdout|int !=2"
+  changed_when: false
+
 
 
 -name: "Verify that Ceilometer Events Smartgateway shows two containers"
@@ -39,6 +47,8 @@
       oc get pods | grep default-cloud1-ceil-event| wc -l
   register: output
   failed_when: "output.stdout|int !=2"
+  changed_when: false
+
 
 
 -name: "Verify that Ceilometer Metrics  Smartgateway shows two containers"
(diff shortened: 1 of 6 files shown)


roles/test_ha/tasks/main.yml
```diff
@@ -0,0 +1,90 @@
+---
+# Assuming we've already logged in...
+# Following procedure on https://infrawatch.github.io/documentation/#high-availability_assembly-advanced-features
+# Assuming we're in teh right project already...
+
+- name: "Add highAvailability.enabled: true to the spec section"
+  ansible.builtin.shell:
+    cmd: |
+      oc patch stf/default --type merge -p '{"spec":{"highAvailability":{"enabled": true}}}' ; sleep 120
```
csibbitt, line 9: I think this might not fail correctly if the `oc patch` fails for some reason, because the exit status will get swallowed and we'll get the `sleep` exit status instead. This failure would cause future tests to fail, but the root cause could be obscured. Suggest using `&&` instead of `;` (untested)

csibbitt, line 9:
```suggestion
      oc patch stf/default --type merge -p '{"spec":{"highAvailability":{"enabled": true}}}' && sleep 120
```

roles/test_ha/tasks/main.yml
```diff
@@ -0,0 +1,90 @@
...
+  changed_when: false
```
csibbitt, line 10: I think all of the shell blocks could benefit from this.

roles/test_snmpTraps/tasks/main.yml
```diff
@@ -15,12 +15,6 @@
       for i in {1..3}; do oc delete po -l application=default-interconnect;  done
   changed_when: false
 
-- name: "Check for snmpTraps logs"
```
csibbitt, line 18: Make sure to double check that these SNMP tests still pass after this change, since you're decreasing the wait time by ~77% here.

The command itself actually looks suspect to me - I would expect to put a sleep in there to standardize the wait length, otherwise isn't it basically dictated by the speed of the API response? The examples from our "STF Feature Verification" wiki use `sleep 1`.

elfiesmelfie (commented):
Besides Chris's comments, it looks okay to me.
The READMEs need to be updated, but we've merged the others so far.

roles/test_snmpTraps/tasks/main.yml
lnatapov: Done

lnatapov:
> Besides Chris's comments, it looks okay to me. The READMEs need to be updated, but we've merged the others so far.

I made a changes

elfiesmelfie (approved):

roles/test_snmpTraps/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
 - name: "Interrupt metrics flow by preventing the QDR"
   ansible.builtin.shell:
     cmd: |
-      for i in {1..3}; do oc delete po -l application=default-interconnect;  done
+      for i in {1..3}; do oc delete po -l application=default-interconnect; sleep 1; done
```
csibbitt, line 15:
```suggestion
      for i in {1..30}; do oc delete po -l application=default-interconnect; sleep 1; done
```

roles/test_snmpTraps/tasks/main.yml
csibbitt: I'm still skeptical of this test and don't think it will be reliable after this change (but it could have been reliable before the PR since it ran 10 extra times)

The default alertmanager route we describe in the docs requires 30s of disruption at minimum before it will send an alert. We don't want to sleep for long because the pod may be able to restart and transmit some data, so we should keep the `sleep 1` but increase the loop size to guarantee that this lasts at least 30s. I have proposed what I think is the best solution (though it's going to run much longer than even the old code before this PR)

lnatapov:
@csibbitt Thank,Chris. I will change that.

PR #40, opened 2023-02-14
Adding observability Strategy tests and fixing typo in Prometheus URL variable in test_alerts

lnatapov:


roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+    cmd: |
+      oc patch stf/default --type merge -p '{"spec":{"observabilityStrategy": "none"}}' && sleep 120
+  changed_when: false
+
+
+- name: "Delete Grafana object"
+  ansible.builtin.shell:
+    cmd: |
+      oc delete grafana default && sleep 60
```
csibbitt, line 17: These sleeps make me a bit nervous because they are unevenly applied. For example, the time between "Delete Grafana object" and "Verify that Grafana object didn't respawn" is >= 4 mins, whereas the time between  "Delete ElasticSearch object" and "Verify that elasticsearch object didn't respawn" is `4 * <get pods time>` which hopefully is a lot less than 4 mins.

I think ultimately what you're trying to do here is wait long enough to ensure that the pods actually disappeared, and then additional time to make sure they don't come back. 

One approach that comes to mind is to replace the `sleep 60`s with something like `while oc get po -l app=grafana | grep grafana; do sleep 1; done;`. This would actually ensure the pod was gone, and would usually take less time than the 60s sleep, too. Then you'd be sure that each pod disappeared, and could do a single sleep (even up to that original 4 minutes if you think it's necesarry) before checking to ensure they don't return.

roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+     cmd: |
+       oc get pods | grep event| wc -l
+  register: output
+  failed_when: "output.stdout|int !=0"
+  changed_when: false
+
+
+
+- name: "Verify that Grafana object didn't respawn"
```
csibbitt, line 52:
```suggestion
- name: "Verify that Grafana pod didn't respawn"
```

roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+     cmd: |
+       oc get pods | grep grafana-deployment | wc -l
+  register: output
+  failed_when: "output.stdout|int !=0"
+  changed_when: false
+
+
+- name: "Verify that Prometheus object didn't respawn"
```
csibbitt, line 61:
```suggestion
- name: "Verify that Prometheus pod didn't respawn"
```

roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+     cmd: |
+       oc get pods | grep prometheus-default | wc -l
+  register: output
+  failed_when: "output.stdout|int !=0"
+  changed_when: false
+
+
+- name: "Verify that Alertmanager object didn't respawn"
```
csibbitt, line 70:
```suggestion
- name: "Verify that Alertmanager pod didn't respawn"
```

roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+     cmd: |
+       oc get pods | grep alertmanager-default | wc -l
+  register: output
+  failed_when: "output.stdout|int !=0"
+  changed_when: false
+
+
+- name: "Verify that elasticsearch object didn't respawn"
```
csibbitt, line 79:
```suggestion
- name: "Verify that elasticsearch pod didn't respawn"
```

roles/test_observabilityStrategy/tasks/main.yml
csibbitt: Suggesting these description changes to help distinguish between the object that the operator acts on, and the pods that are a result of the operator's actions.

roles/test_observabilityStrategy/tasks/main.yml
lnatapov: Thanks for the comment. I made the changes.

roles/test_observabilityStrategy/tasks/main.yml
```diff
@@ -7,35 +7,42 @@
...
 - name: "Delete ElasticSearch  object"
   ansible.builtin.shell:
     cmd: |
-      oc delete elasticsearches.elasticsearch.k8s.elastic.co elasticsearch && sleep 60
+      oc delete elasticsearches.elasticsearch.k8s.elastic.co elasticsearch &&  while oc get po | grep elasticsearch; do sleep 1; done;
+  changed_when: false
+
+
+- name: "Wait 2 minutes before checking that pods didn't return"
+  ansible.builtin.shell:
```
csibbitt, line 43: There is an ansible module for this if you prefer https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pause_module.html

csibbitt (approved):
LGTM

roles/test_observabilityStrategy/tasks/main.yml
lnatapov: Nice !

csibbitt (approved):

PR #42, opened 2023-03-02
Fixing indentation issue with HA tests

lnatapov:
Fixing indentation issue with HA tests
Adding pause module to HA tests


elfiesmelfie (approved):
I'm not sure about the sleep 120, but we can figure out a better solution to that later.
Otherwise, this looks good to me

PR #50, opened 2023-04-20
Adding role for collectd sensubility

lnatapov:
1.Adding test for container down health status
2.Adding test for container up health status

### DIFF
--- a/roles/test_sensubility/README.md
+++ b/roles/test_sensubility/README.md
@@ -1,0 +1,38 @@
+test_ha
+==============
+
+Tests for Collectd Sensubility
+
+Requirements
+------------
+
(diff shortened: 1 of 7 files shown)


elfiesmelfie (commented):
I think this could be grouped with collectd.

I can refactor to make test_collectd into its own role.

lnatapov:
> I think this could be grouped with collectd.
> 
> I can refactor to make test_collectd into its own role.

It makes sense

elfiesmelfie:
I'm working on refactoring the collectd (and other roles) here: https://github.com/infrawatch/functional-tests/pull/53

csibbitt (approved):
LGTM

PR #57, opened 2023-10-09
Functional alexy

ayefimov-1:
Please review this change.   I added a new inv file that should be used specifically for the cvp 
 compact jobs.  The only difference from the default.inv is in the last line
 
https://github.com/infrawatch/functional-tests/compare/functional-alexy?expand=1#diff-81f746bf052c8428eedaa824c7fc262bdc4290d0b751239acb2ea1a0ac44140d


lnatapov (commented):
This inv file is tailored for long lived STF on PSI. Note Line 52: 

ocp_cluster ansible_host="10.0.77.132" ansible_user="cloud-user" ansible_ssh_private_key_file="jumphost.key"

ayefimov-1:
> This inv file is tailored for long lived STF on PSI. Note Line 52:
> 
> ocp_cluster ansible_host="10.0.77.132" ansible_user="cloud-user" ansible_ssh_private_key_file="jumphost.key"

@lnatapov this is intended to be long lived. The location of the jumphost.key file is slightly different in compact jobs.

lnatapov (commented):
This is going to run only against 10.0.77.130. We have talked about variable there {{ocp_ip}}.

ayefimov-1:
Final update to default-compact.inv file.

elfiesmelfie:
I would advise not adding the new tasks to get_prom_info.yml

Before you run the functional tests, you should log into the cluster.

default-compact.inv
```diff
@@ -0,0 +1,51 @@
+# NOTE: for defining hosts both hostnames and IP addresses forms can be used
```
elfiesmelfie, line 1: All these lines are commented, so IMHO, this doesn't serve a purpose except as a sample inv file, in which case, the name should be changed

roles/client_side_tests/tasks/get_prom_info.yml
```diff
@@ -1,4 +1,17 @@
 ---
+- name: "Get kube password"
+  ansible.builtin.shell:
+    cmd: |
```
elfiesmelfie, line 4: This is very specific to the installer used to provision openshift.

This won't work, for example, on a cluster created with CRC.

I'd prefer to keep the login part as a pre-task that the user is expected to perform before running the functional tests.

PR #59, opened 2023-11-16
Add Functional test for Dashboards

mgirgisf:


roles/test_dashboards_loaded/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+  ansible.builtin.shell:
+    cmd: |
+      oc apply -f https://raw.githubusercontent.com/infrawatch/dashboards/master/deploy/stf-1/rhos-dashboard.yaml
+      oc apply -f https://raw.githubusercontent.com/infrawatch/dashboards/master/deploy/stf-1/virtual-machine-view.yaml
+      oc apply -f https://raw.githubusercontent.com/infrawatch/dashboards/master/deploy/stf-1/memcached-dashboard.yaml
+  changed_when: false
+  register: cmd_output
+  failed_when: cmd_output.rc != 0
+
+- name: "Verify Dashboards"
```
csibbitt, line 76: I'm not very convinced by this test. All you're actually testing is the `oc apply` command above which implies that the Grafana Operator CRDs got installed far enough that the grafanadashboards API is available. Neither the Grafana Operator nor STF needs to be functioning in order to pass this test, and it doesn't actually test if the dashboards are "loaded" into Grafana (as implied by the role name), let alone "functioning" (as implied by the project name).

We could get a bit closer by curl'ing Grafana itself to make sure that the dashboards are appearing in the UI. This would prove:
1.  The Grafana Operator and Grafana itself are actually functioning beyond creating a route (though this is testing Grafana Operator more than STF or dashboards, it's still useful)
3. The dashboard json is valid enough to get loaded successfully into Grafana. This would be a large step forward vs simply checking that the grafanadashboard object was valid enough to be loaded into OCP. To me this seems like the minimum.

Valid json still doesn't validate the dashboard _function_ or that STF is populating those dashboards. If I could dream (and it's definitely asking a lot for an initial scope, so please treat this as an aspirational stretch goal or future work), I'd like to curl the dashboards themselves and ensure that every panel actually has data series' being displayed. This would additional prove:
1. The dashboards are written to match the metrics being delivered
2. STF is delivering all of the metrics required to populate the dashboards.

stf_functional_tests.yml
```diff
@@ -62,6 +62,11 @@
       ansible.builtin.import_role:
         name: test_snmp_traps
 
+
+    - name: " Test Loaded Dashboards"
```
csibbitt, line 66:
```suggestion
    - name: "Test Loaded Dashboards"
```

roles/test_dashboards_loaded/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+        name: grafana-operator
+        source: community-operators
+        sourceNamespace: openshift-marketplace
+      EOF
+
+  changed_when: false
+  register: cmd_output
+  failed_when: cmd_output.rc != 0
+
+- name: "Manual Approve installplan"
```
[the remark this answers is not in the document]
mgirgisf: it didn't work in my env without it so I wasn't sure if it may cause a problem, I know its an extra step that may not be needed but it takes seconds in running and not causing any conflicts , but I can try to test it again on another fresh env and check .

roles/test_dashboards_loaded/tasks/main.yml
mgirgisf: okay you are right , its better to check with  curl'ing Grafana itself , I will update this part starting with valid json and adding this extra validation for each dashboards in jira to work on it as extra testing

PR #60, opened 2023-11-17
Test for Email Alert

mgirgisf:


roles/test_verify_email/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+- name: "Patch the ServiceTelemetry object for the STF deployment"
+  ansible.builtin.shell:
+    cmd: |
+      oc patch stf default --type merge -p '{"spec": {"alertmanagerConfigManifest": "apiVersion: v1\nkind: Secret\nmetadata:\n  name: 'alertmanager-default'\n  namespace: 'service-telemetry'\ntype: Opaque\nstringData:\n  alertmanager.yaml: |-\n    global:\n      resolve_timeout: 10m\n    route:\n      group_by: ['job']\n      group_wait: 30s\n      group_interval: 5m\n      repeat_interval: 12h\n      receiver: 'email'\n    receivers:\n    - name: 'email'\n      - email_configs:\n      - to: mx@redhat.com"}}'
+  changed_when: false
+
+- name: "Interrupt metrics flow by preventing the QDR from running"
+  ansible.builtin.shell:
+    cmd: |
+      for i in {1..30}; do oc delete po -l application=default-interconnect; sleep 1; done
```
csibbitt, line 37: I think I may be the originator of this technique, and should mention that I don't think it's entirely reliable; I swear I've seen the QDR start up and pass some metrics within 1s, but it's rare.

Here though, I'm a bit extra skeptical because you have ~30s worth of interruption, but the alert query is looking for a rate of 0 metrics over a 60s time frame. I would guess that if there were actually metrics flowing, this interruption would not be long enough to trigger the alert. Since the test is meant to test the email alert function (and related rule/config documentation), I think it's perfectly valid to test it in a quiet system with no metrics and no simulated interruption, but want to point out that we could get false-negatives here for a couple related reasons if the system is not quiet.

roles/test_verify_email/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+
+- name: "Check for alertmanager logs"
+  ansible.builtin.shell:
+    cmd: |
+      oc logs alertmanager-default-0 -c alertmanager  | grep 'receiver=email' | wc -l
+  register: cmd_output
+  failed_when: "cmd_output.stdout|int == 0"
+  changed_when: false
+
+- name: "Wait 2 minutes to make sure all SG pods are back to normal"
```
csibbitt, line 48: What is this? I don't see anything happening to the SG pods previous to this, and you've already parsed success from the alertmanager log - what exactly are you waiting for here?

PR #62, opened 2023-12-05
Use OAuth token instead of http basic auth for prometheus access

csibbitt:
This change is required in order to match changes in STO where we will stop allowing http basic auth for access from grafana to prometheus, and use an oauth token from a new restricted service account instead.

Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/549/

### DIFF
--- a/roles/client_side_tests/tasks/get_prom_info.yml
+++ b/roles/client_side_tests/tasks/get_prom_info.yml
@@ -1,14 +1,14 @@
 ---
-- name: "Get the default-prometheus-htpasswd secret"
+- name: "Generate an access token for prometheus"
   ansible.builtin.shell:
     cmd: |
-      oc  get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
-  register: prom_secret
+      oc create token stf-prometheus-reader
+  register: prom_token_out
   changed_when: false
 
 - name: "Show the prom_secret value"
   ansible.builtin.debug:
-    var: prom_secret | string
+    var: prom_token_out | string
 
 - name: "Get the prom URL"
   ansible.builtin.shell:
@@ -23,16 +23,5 @@
 
 - name: "Get the prom creds from the secret"
   ansible.builtin.set_fact:
-    prom_user_decoded: "{{ (prom_secret.stdout_lines[0] | b64decode) }}"
-    prom_pass: "{{ prom_secret.stdout_lines[1] | b64decode }}"
+    prom_token: "{{ prom_token_out.stdout }}"
     prom_url: "{{ prom_route.stdout }}"
-
-- name: "Show the prom_user value"
-  ansible.builtin.debug:
-    var: prom_user
-
-- name: "Fetch user"
-  ansible.builtin.set_fact:
-    prom_user: "{{ prom_user_decoded.split(':')[0] }}"
-  when: prom_user_decoded is defined
-
--- a/roles/client_side_tests/tasks/get_prom_info.yml
+++ b/roles/client_side_tests/tasks/get_prom_info.yml
@@ -1,14 +1,47 @@
 ---
-- name: "Generate an access token for prometheus"
-  ansible.builtin.shell:
-    cmd: |
-      oc create token stf-prometheus-reader
-  register: prom_token_out
-  changed_when: false
+- when: prom_auth_method == 'password'
+  block:
+  - name: "Get the default-prometheus-htpasswd secret"
+    ansible.builtin.shell:
+      cmd: |
+        oc  get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
+    register: prom_secret
+    changed_when: false
 
-- name: "Show the prom_secret value"
-  ansible.builtin.debug:
-    var: prom_token_out | string
+  - name: "Show the prom_secret value"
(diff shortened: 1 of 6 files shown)


csibbitt:
NOTE: This change will not work until  https://github.com/infrawatch/service-telemetry-operator/pull/549/ merges, and will not work with existing STF releases!

I tested this locally against an environment deployed from my STO PR branch like so: `ansible localhost --module-name include_role --args name=client_side_tests` and was able to see successful prometheus queries using the token.

Before we merge this, I'm noticing that we may have a problem. While this change is required in order to test the new code that will soon be on the master branch, there are no "stable-1.5"-like release branches in this repo, so if this merges, you will no longer be able to use it to test the release version of STF. Creating the branch is trivial, but I have not yet looked at what will be required to adjust the jenkins jobs to use the correct branch at the correct time, and feel like this is way outside my scope at the moment. Looking for help from QE to figure out how to handle this.

CC: @ayefimov-1 @myadla

elfiesmelfie:
> NOTE: This change will not work until [infrawatch/service-telemetry-operator#549](https://github.com/infrawatch/service-telemetry-operator/pull/549) merges, and will not work with existing STF releases!
> 
> I tested this locally against an environment deployed from my STO PR branch like so: `ansible localhost --module-name include_role --args name=client_side_tests` and was able to see successful prometheus queries using the token.
> 
> Before we merge this, I'm noticing that we may have a problem. While this change is required in order to test the new code that will soon be on the STO master branch, there are no "stable-1.5"-like release branches in this repo, so if this merges, you will no longer be able to use it to test the release version of STF. Creating the branch is trivial, but I have not yet looked at what will be required to adjust the jenkins jobs to use the correct branch at the correct time, and feel like this is way outside my scope at the moment. Looking for help from QE to figure out how to handle this.
> 
> CC: @ayefimov-1 @myadla

I would suggest adding a var to determine whether to use user/pass or get a token.
prom_auth_method = token|pass

Also, a branch would be a good idea, but we still have a temporary issue trying to figure out how to point jenkins at the right branch.

roles/client_side_tests/tasks/get_prom_info.yml
```diff
@@ -1,14 +1,14 @@
 ---
-- name: "Get the default-prometheus-htpasswd secret"
+- name: "Generate an access token for prometheus"
   ansible.builtin.shell:
     cmd: |
-      oc  get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
-  register: prom_secret
+      oc create token stf-prometheus-reader
```
elfiesmelfie, line 5: For backwards-compatibility, I suggest keeping the "get password" and "get user" tasks, and stick them in a block, guarded with a `when: X`.

The new token generation can be in a complementary block (i.e. `when: not X`)

To keep the curl tasks clean, I suggest adding a "prom_auth_string" that would be either `-u "{{ prom_user }}:{{ prom_pass }}"` or `-u "{{ prom_user }}:{{ prom_pass }}"`

csibbitt:
Thanks, @elfiesmelfie , I'll add a variable that selects between the methods as a temporary measure until the branch awareness can be sorted out.

roles/client_side_tests/tasks/get_prom_info.yml
```diff
@@ -1,14 +1,47 @@
 ---
-- name: "Get the default-prometheus-htpasswd secret"
-  ansible.builtin.shell:
-    cmd: |
-      oc  get secret default-prometheus-htpasswd -ojson | jq '.data.auth, .data.password' |  sed 's/"//g'
-  register: prom_secret
-  changed_when: false
+- when: prom_auth_method == 'password'
+  block:
+  - name: "Get the default-prometheus-htpasswd secret"
```
csibbitt, line 4: This diff is unfair; every one of these blocks was in the existing code! ;) The changes are mostly indenting.

csibbitt:
I haven't looked into why, but the task names are never printed during execution so it's hard to follow; nevertheless, here is my evidence of testing:

### Token auth
STF  from csibbitt/STF-1549_prometheus_token_auth branch
```
$ ansible localhost --module-name include_role --args name=client_side_tests
[...]
localhost | SUCCESS | rc=0 >>
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   136    0    63  100    73    431    500 --:--:-- --:--:-- --:--:--   931
```
### Password auth
STF from stable-1.5 branch

Fails with token auth:
```
$ ansible localhost --module-name include_role --args name=client_side_tests
[...]
localhost | FAILED | rc=1 >>
error: failed to create token: serviceaccounts "stf-prometheus-reader" not foundnon-zero return code
```

Works with password auth:
```
$ ansible localhost --module-name include_role --args name=client_side_tests -e prom_auth_method=password
[...]
localhost | SUCCESS | rc=0 >>
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   136    0    63  100    73    140    162 --:--:-- --:--:-- --:--:--   302
```

When this merges, I'll create a stable-1.5 branch in this repo and set the prom_auth_method to "password" on that branch; that way we can have both this code-heavy workaround and a future-proof branching model all rolled into one!

ayefimov-1 (approved):
Looks fine to me.

PR #73, opened 2024-03-11
[zuul] Run tests on OSP18 

mgirgisf:


elfiesmelfie:
This looks good!
The output status of the tests is not indicative of passing tests though.
The functional_test playbook uses ignore_errors: true, so that all the tests can be done.
We use this in  a downstream job, and there is an output plugin ([callback_plugins/custom_logger.py](https://github.com/infrawatch/feature-verification-tests/blob/master/callback_plugins/custom_logger.py)) configured to report the passs/fail rates. 

Can you add a more verbose output?
And put the logs in one of the directories that are automatically uploaded to Zuul?

.zuul.yaml
```diff
@@ -1,6 +1,27 @@
 ---
+- job:
+    name: stf-functional-tests-on-osp18
+    parent: telemetry-operator-multinode-autoscaling
+    description: |
+      Run the stf functional tests on osp18
+    required-projects:
```
elfiesmelfie, line 7: If you want to re-use this job in another repo, then the required projects should be updated to include this repo

here is an example:  
https://github.com/infrawatch/service-telemetry-operator/blob/853e4574b5d3c93faa2b9001f8bf18f0a9eddc9e/.zuul.yaml#L90
The stf jobs would not run on the other infrawatch repos until I added STO to the required-projects.

elfiesmelfie (changes requested):

morucci:
recheck

ci/zuul_stf_tests.yml
```diff
@@ -0,0 +1,92 @@
...
+        - name: "Check for podman executable"
+          ansible.builtin.stat:
+            path: "/usr/bin/podman"
+          register: st_pod
+        - name: "Set container_bin=docker when podman isn't there"
+          ansible.builtin.set_fact:
+            container_bin: docker
+          when: not st_pod.stat.exists
+
+    - name: "Run collectd tests"
```
elfiesmelfie, line 27: For OSP18. this is not required

ci/collect_logs.yml
```diff
@@ -0,0 +1,38 @@
...
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/zuul-output/logs"
+        state: directory
+        mode: "0755"
+
+- hosts: localhost
+  name: Collect logs on the controller
+  gather_facts: false
+  tasks:
+    - name: Create log dir
```
elfiesmelfie, line 16: I don't know how much Zuul likes things being created on the executer node.

It seems that this does work, so perhaps we can update the output_dir for the custom_logger to put the files directly into the {{ ansible_user_dir }}/zuul-output/logs/controller/ dir.

This needs some investigation, i.e. is the output dir hardcoded in the plugin, and how can it be updated?

ci/collect_logs.yml
```diff
@@ -0,0 +1,38 @@
...
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/zuul-output/logs/controller"
+        state: directory
+        mode: "0755"
+
+    - name: Copy ansible logs
+      ansible.builtin.shell: |
+        cp {{ ansible_env.HOME }}/test_run_result.out .
```
elfiesmelfie, line 24: test_run_result.out actually isn't in $HOME. It's in playbook_dir by default.

I think you should be able to just reference {{ playbook_dir }} since all the ci/ playbooks are in teh same place.

ci/pre-2node.yml
elfiesmelfie, line 1: This is not needed, since the pre-run playbooks for the parent jobs will include this.

I suspect that the issue is that you're trying to run against all hosts, which includes `crc`, which doesn't have the kubeconfig file created (by design).

ci/run.yml
```diff
@@ -0,0 +1,33 @@
...
+        chdir: "{{ fvt_dir }}"
+      register: output
+
+    - name: Save the output of the playbook run to a logfile
+      ansible.builtin.lineinfile:
+        path: "{{ ansible_env.HOME }}/stf_test.log"
+        line: "{{ output.stdout_lines }}"
+        create: yes
+
+    - name: "Run a playbook to gather logs on the target host"
```
elfiesmelfie, line 26: You can run collect logs as normal, in a post-run step

mgirgisf:
recheck

ci/ansible.cfg
```diff
@@ -0,0 +1,5 @@
+[defaults]
+callback_whitelist = custom_logger
+# additional paths to search for roles 
+roles_path = ~/feature-verification-tests/roles:./roles:/usr/share/ansible/roles:/etc/ansible/roles:~/.ansible/roles:/home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles
```
elfiesmelfie, line 4:
```suggestion
roles_path = /home/zuul/src/github.com/infrawatch/feature-verification-tests/roles:/usr/share/ansible/roles:/etc/ansible/roles:~/.ansible/roles
```

ci/pre-2node.yml
elfiesmelfie, line 1: This file should not be needed if you're using the telemetry-operator-multinode-autoscaling as the parent job.

ci-framework/playbooks/01-bootstrap.yml and 02-infra.yml are already called by the pre-run steps in the parent jobs.
`sto_dir` is not needed.

You may need to pass in a var that points at the kubeconfig file to the appropriate playbooks.

ci/stf_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        - name: "Check for podman executable"
+          ansible.builtin.stat:
+            path: "/usr/bin/podman"
+          register: st_pod
+        - name: "Set container_bin=docker when podman isn't there"
+          ansible.builtin.set_fact:
+            container_bin: docker
+          when: not st_pod.stat.exists
+
+    - name: "Run collectd tests"
```
elfiesmelfie, line 27: We don't deploy collectd for 18, so this can be removed.

Unless you want some conditional execution depending on the version of openstack being used.
(I would recommend keeping conditional execution in mind, when making design decisions but not actually adding in anything conditional until the OSP17 job gets added)

ci/ansible.cfg
mgirgisf: I needed openstack-k8s-operators/ci-framework/roles for running  the task cluster login from the role : rhol_crc

ci/stf_tests.yml
mgirgisf: I'm to getting the list of required test for osp18 from leonid and I will remove all the unnecessary tasks untill I created the osp17 job

ci/ansible.cfg
elfiesmelfie: Of course. I saw that and didn't join the dots.

elfiesmelfie:
> Merge Failed.
> 
> This change or one of its cross-repo dependencies was unable to be automatically merged with the current state of its repository. Please rebase the change and upload a new patchset. Warning: Error merging github.com/infrawatch/feature-verification-tests for 73,79062f7b9a2a0ca66ee8a9fbcc6ccb157245cc0e

@mgirgisf I've resolved the conflicts with main. Let me know if it doesn't work as expected.

ci/run.yml
```diff
@@ -0,0 +1,30 @@
...
+        name:
+          - ansible
+
+    - name: Set the repo dir
+      ansible.builtin.set_fact:
+        fvt_dir: "{{ ansible_env.HOME}}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
+
+    - name: Copy the callback_plugin to enable it
+      ansible.builtin.copy:
+        src: "{{ fvt_dir }}/callback_plugins/"
```
elfiesmelfie, line 14: This is failing.

This is because of the copy module. The `src` is on the ansible deployment host i.e. zuul executer or localhost.

What you're expecting is the equivalent of `cp fvt/callback_plugins fvt/ci/callback_plugins`

What's actually happening is closer to `scp zuul_executer:/fvt/callback_plugin controller:fvt/ci/callback_plugins`

You need to use the [`remote_src`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html#parameter-remote_src) option to specify that the file is already present on the target host, and to copy from file_on_target to file_on_target.

ci/run.yml
```diff
@@ -0,0 +1,30 @@
...
+        dest: "{{ fvt_dir }}/ci/callback_plugins/"
+
+    - name: "Run a playbook locally on the target host"
+      ansible.builtin.shell:
+        cmd: |
+          ANSIBLE_CONFIG=ansible.cfg ansible-playbook stf_tests.yml
+        chdir: "{{ fvt_dir }}/ci"
```
elfiesmelfie, line 21: I think that if you change this to ``chdir: "{{ fvt_dir }}"``, and `... ansible-playbook ci/stf_tests.yml`, that you don't need to move the callback_plugin dir.

Alternatively, you can tell ansible where to look for callback plugins, the same way that you tell ansible where to look for roles.
https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-callback-plugin-path

ci/roles/client_side_tests/tasks/test_e2e.yml
elfiesmelfie, line 1: You need to sync with infrawatch/feature-verification-tests:master to get rid of these changes. They're not part of your update, are they?

ci/run.yml
mgirgisf: fo some reason ts not working as For ansible to use the callback_plugins directory it must be located beside the playbook, in the same directory. we need to have something like that, I'm playing around to find a way.

├── callback_plugins
│   └── env_warning.py
└── playbook.yml

ci/roles/client_side_tests/tasks/test_e2e.yml
mgirgisf: No they are, I made a copy to make changes for osp18 for this role without affecting the main one.

mgirgisf:
recheck

.zuul.yaml
```diff
@@ -1,6 +1,36 @@
...
+        override-checkout: main
+      - name: github.com/openstack-k8s-operators/dataplane-operator
+        override-checkout: main
+      - name: github.com/openstack-k8s-operators/infra-operator
+        override-checkout: main
+      - name: github.com/openstack-k8s-operators/openstack-operator
+        override-checkout: main
+      - name: github.com/openstack-k8s-operators/openstack-must-gather
+        override-checkout: main
+      - name: github.com/openstack-k8s-operators/telemetry-operator
```
csibbitt, line 21: This is the replacement for STF - are we intending to use this code to actually test Telemetry Operator?

.zuul.yaml
```diff
@@ -1,6 +1,36 @@
 ---
+- job:
+    name: functional-tests-on-osp18
+    parent: telemetry-operator-multinode-autoscaling
+    description: |
+      Run the stf functional tests on osp18
```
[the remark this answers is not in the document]
mgirgisf: As its still draft, is the main idea here to be able to run some tests with osp18 specially autoscaling test, but autoscaling test is still in progress , so I'm working on the zuul part to be able to run ansible tests in general, gather the logs and gather the result of callback_plugins , so when the autoscaling test is ready we are able to just add it with working ci job.

ci/roles/client_side_tests/tasks/test_e2e.yml
```diff
@@ -0,0 +1,121 @@
+---
```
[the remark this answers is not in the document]
mgirgisf: its the same exactly I didn't change anything to it yet, I just had a copy from one role to play around  if I need to change something, but when this PR is ready I will remove it, but for now I'm working on enable the callback_plugin and just complete the CI .

mgirgisf:
recheck

ci/stf_tests.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: "Run e2e tests"
+  hosts: localhost
+  ignore_errors: true
+  environment:
+    PATH: "~/.crc/bin:~/.crc/bin/oc:~/bin:{{ ansible_env.PATH }}"
+    cifmw_rhol_crc_kubeconfig: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"
```
elfiesmelfie, line 7: This most likely should be

```suggestion
    KUBECONFIG: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"
```

.zuul.yaml
elfiesmelfie: We're going to add more tests as well, and use these feature-verification tests for both telemetry-operator and tripleo deployments.

ci/run.yml
```diff
@@ -0,0 +1,25 @@
...
+      ansible.builtin.pip:
+        name:
+          - ansible
+
+    - name: "Run a playbook locally on the target host"
+      ansible.builtin.shell:
+        cmd: |
+          ANSIBLE_CONFIG=ci/ansible.cfg ansible-playbook -v ci/stf_tests.yml
+        chdir: "{{ fvt_dir }}"
+      register: output
```
elfiesmelfie, line 15: Okay, I think the issue is that the cifmw vars are available in teh context of this run playbook. But the stf_tests.yml playbook is completely isolated from the vars and facts in this context.

We need to pass in the vars we want.
In this case, cifmw_path and cifmw_kubeconfig.

You could do the login here (i.e. ci.run) and then pass in the KUBECONFIG env var to this task. (option 1)
```suggestion
    - name: "Run a playbook locally on the target host"
      ansible.builtin.shell:
        cmd: |
          ANSIBLE_CONFIG=ci/ansible.cfg ansible-playbook -v ci/stf_tests.yml
        chdir: "{{ fvt_dir }}"
      environment:
          KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
          PATH: "{{ cifmw_path }}"
      register: output
```
You could also pass in all the required vars here. and do the login in stf_tests. (option 2)

However, I would recommend option 1. We can then use the feature-verification tests assuming that we're already logged in, and then we can keep the cifmw vars separate form the role config vars.

ci/stf_tests.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: "Run e2e tests"
+  hosts: localhost
+  ignore_errors: true
+  gather_facts: false
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
```
elfiesmelfie, line 8: Going with option 1 above, you also need to remove these

```suggestion
```

ci/run.yml
```diff
@@ -0,0 +1,25 @@
+- hosts: controller
+  vars:
+    fvt_dir: "{{ ansible_env.HOME}}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
+  tasks:
+    - name: Install ansible so that ansible-playbook is available
+      ansible.builtin.pip:
+        name:
+          - ansible
+
```
elfiesmelfie, line 9: Add in a login command here. Either using rhol_crc or `command: cmd='oc login -u {{ cifmw_openshift_user }} -p {{ cifmw_openshift_password }} {{ cifmw_openshift_api }}'`

Code suggestions should be audited to update format, and make sure they're correct

elfiesmelfie (changes requested):

elfiesmelfie:
The job passed, but it wasn't really  pass.


2024-04-24 08:40:36.737301 | controller | TASK [try to run oc command] ***************************************************
2024-04-24 08:40:36.764744 | controller | fatal: [localhost]: FAILED! => {"msg": "The field 'environment' has an invalid value, which includes an undefined variable. The error was: 'cifmw_openshift_kubeconfig' is undefined. 'cifmw_openshift_kubeconfig' is undefined\n\nThe error appears to be in '/home/zuul/src/github.com/infrawatch/feature-verification-tests/ci/stf_tests.yml': line 10, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n    - name: \"try to run oc command\"\n      ^ here\n"}
2024-04-24 08:40:36.764826 | controller | ...ignoring

elfiesmelfie:
recheck

elfiesmelfie:
recheck

ci/run.yml
```diff
@@ -0,0 +1,38 @@
+- hosts: controller
+  vars:
+    fvt_dir: "{{ ansible_env.HOME}}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
```
elfiesmelfie, line 6:
```suggestion
    PATH: "{{ cifmw_path if cifmw_path is defined else ['~/.crc/bin:~/.crc/bin/oc:~/bin', ansible_env.PATH ] | join(':') }}"
```

elfiesmelfie: I highly recommend using this workaround for now, and figuring out later where the cifmw_path is set, once the rest is working.

PR #80, opened 2024-04-26
Update callback plugin

mgirgisf:
Update custom_logger to generate summary file called "test_run_result.out" with the number of failures and success and then generate a detailed report for each host in separate file called "{host}_results.log"

### DIFF
--- a/callback_plugins/custom_logger.py
+++ b/callback_plugins/custom_logger.py
@@ -7,34 +7,41 @@ class CallbackModule(CallbackBase):
     CALLBACK_TYPE = 'notification'
     CALLBACK_NAME = 'log_to_file'
     CALLBACK_NEEDS_WHITELIST = True
+    MSG_FORMAT = "{task}={status}\n"
 
     def __init__(self):
         super(CallbackModule, self).__init__()
         self.output_dir = os.getcwd()
         self.results = {}
-
+        
     def playbook_on_stats(self, stats):
         # Log results for each host
         hosts= stats.processed
         for host in hosts:
-            self._log_results(host)
-        self._summary_results(hosts)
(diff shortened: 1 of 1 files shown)


callback_plugins/custom_logger.py
```diff
@@ -1,72 +1,69 @@
...
-    def log(self, host, category):
-         # only interested in the test tasks, not setup or debug
-         if self.current_task.startswith("RHELOSP"):
-
-             if category == "failed":
-                self.failed +=1
-             if category == "passed":
-                 self.passed +=1
-
-             test_result_message = self.MSG_FORMAT.format(task=self.current_task, status=category)
```
elfiesmelfie, line 56: There's a requirement on the format of the results.out file to remain consistent.
The format needs to be "{task_name}={passed|failed}"

elfiesmelfie: The format should be {testID}={passed|failed}

Existing users expect the test_run_result.out file to contain results in this format.

mgirgisf: I kept the file "test_run_result.out" using the same old format , I just updated the functions to be more organised and we will have another file called "{host}_results.log" that will have more details so we can use it in the zuul job.

lnatapov (approved):

mgirgisf:
**Depends-On:** #90

elfiesmelfie:
> **Depends-On:** #90

The Depends-On needs to be in the PR description -- I've updated it.
I don't know whether the GHA tests will run or not, since it's Zuul that acts on Depends-On.

elfiesmelfie (changes requested):

PR #82, opened 2024-05-14
[telemetry_autoscaling] Add telemetry_autoscaling role and zuul job

elfiesmelfie:
The role is imported from the infrawatch/telemetry_autoscaling repo. Moving it here will let us take advantage of zuul testing

### DIFF
--- a/ci/telemetry_autoscaling.yml
+++ b/ci/telemetry_autoscaling.yml
@@ -3,6 +3,7 @@
   hosts:  "{{ cifmw_target_hook_host | default('localhost')  }}"
   gather_facts: true
   vars:
+    fvt_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
     ci_framework_logs: "{{ ansible_env.HOME }}/ci-framework-data/logs"
     openstack_cmd: "oc rsh openstackclient openstack"
     metrics_backend: "prometheus"
--- a/ci/telemetry_autoscaling.yml
+++ b/ci/telemetry_autoscaling.yml
@@ -1,0 +1,94 @@
+---
+- name: Telemetry autoscaling tests
+  hosts:  "{{ cifmw_target_hook_host | default('localhost')  }}"
+  gather_facts: true
+  vars:
+    ci_framework_logs: "{{ ansible_env.HOME }}/ci-framework-data/logs"
+    openstack_cmd: "oc rsh openstackclient openstack"
+    metrics_backend: "prometheus"
+    stack_image: "cirros"
+    stack_flavor: "m1.small"
+    stack_network: "private"
+    stack_external_network: "public"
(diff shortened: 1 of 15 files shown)


elfiesmelfie:
recheck

mgirgisf:
recheck

ci/autoscaling_osp18.yaml
mgirgisf, line 1: This file is no longer used

ci/functional_tests.yml
mgirgisf, line 1: This file for upcoming functional tests integration.
we can remove it for now also.

ci/vars-functional-test.yml
```diff
@@ -0,0 +1,7 @@
+---
+# temp: skip os-net-setup
+cifmw_os_net_setup_config: []
```
mgirgisf, line 3: I think skip os-net is no longer needed as we use post_deploy hook

mgirgisf (commented):
We have a lot of refactor for the code itself, but at least for now we have the main job with autoscaling test

ci/patch.yml
elfiesmelfie, line 1: This should be renamed to openstackversion_autoscaling.yaml, or something similar

ci/telemetry_autoscaling.yml
```diff
@@ -0,0 +1,95 @@
...
+            name: telemetry_autoscaling
+            tasks_from: creating_stack
+          tags:
+            - create
+
+        - import_role:
+            name: telemetry_autoscaling
+            tasks_from: test_autoscaling
+          tags:
+            - test
```
elfiesmelfie, line 50: This should be replaced with just the one import_role task.

When we add the logging tests, we can have a list of functional_tests (defaults to a list for all roles to run for OSP18).
Each of the test roles we want to run will be included here, and each will be guarded with `when: "test_autoscaling" in functional_tests`

```
- import_role:
     name: telemetry_autoscaling
   when: '"telemetry_autoscaling" in functional_tests'

- import_role:
      name: test_logging
   when: '"test_logging" in functional_tests' 
 ```

Or we can use a with_items: <functional_tests>, and import_role: item
e.g. 
```
- import_role:
    name: "{{ item }}"
  with_items: "{{ functional_tests }}"
```

mgirgisf: I was thinking to have a separate playbook for each test , and just call it from the run_playbooks but the idea to have them as a list is much better in scaling and we can rename the telemetry_autoscaling.yml to something more generic to include all the telemetry_tests with osp18

mgirgisf (approved):
the roles need some refactoring and also the vars but we need it to be merged to start refactoring. I approve it for now as it will not cause any conflict with the running job for osp17.

ci/vars-functional-test.yml
elfiesmelfie: I want to test that separately, just to make sure.

PR #87, opened 2024-05-21
[telemetry_autoscaling] Use templates for rendering the heat files

elfiesmelfie:


roles/telemetry_autoscaling/templates/template.yaml.j2
```diff
@@ -0,0 +1,120 @@
...
+              url: {get_attr: [scaleup_policy, signal_url]}
+      query:
+        {% if metrics_backend == "gnocchi" -%}
+        list_join:
+          - ''
+          - - {'=': {server_group: {get_param: "OS::stack_id"}}}
+        {% endif -%}
+        {% if metrics_backend == "prometheus" -%}
+        str_replace:
+          template: "(rate(ceilometer_cpu{resource_name=~'server_name_prefix.*'}[150s]))/10000000"
```
mgirgisf, line 73: maybe its better if we updated the query with the latest change we have in tempest, https://review.opendev.org/c/openstack/telemetry-tempest-plugin/+/920015
there was some failure cases where the rate function returned 0

mgirgisf (changes requested):

roles/telemetry_autoscaling/templates/template.yaml.j2
elfiesmelfie: This follows the documented procedure that Jaromir and Michael shared. 

The queries in TTP are of a different magnitude, with a configurable threshold, and a differing period:

"query": "(rate(ceilometer_cpu{resource_name=~'te-$ENVIRON['RESOURCE_PREFIX'].*'}[$ENVIRON['PROMETHEUS_RATE_DURATION']s])) * 100"

mgirgisf: yes we are using the rate function, but his update cover the duration used by the prometheus.
for stable testing it should be <ceilometer_polling_interval> +<prometheus_scrape_interval>. With other static values  it's possible,that the rate() function will just return 0. and that what happened in downstream with tempest test.

mgirgisf: my suggestion is to have something similar, for example  to have the values as variables with default parameter  we can change if we faced a problem.   

prometheus_rate_duration = (config.CONF.telemetry.ceilometer_polling_interval + config.CONF.telemetry.prometheus_scrape_interval)

    cfg.IntOpt('ceilometer_polling_interval',
               default=300,
               help="Polling interval configured for ceilometer. This can "
                    "be used in test cases to wait for metrics to appear."),
    cfg.IntOpt('prometheus_scrape_interval',
               default=15,
               help="Scrape interval configured for prometheus. This can "
                    "be used in test cases to properly configure autoscaling")

elfiesmelfie: I see. I think this warrants some additional attention and is outside the scope of this change. I'll file a bug for follow-up, so we can also make sure that the documentation covers this information.

mgirgisf (approved):

PR #88, opened 2024-05-22
[ZUUL] Update Variables 

mgirgisf:


mgirgisf:
recheck

ci/vars/osp18_env.yml
elfiesmelfie, line 1: Anything specific to OSP18 should be passed to the OSP18 job IMHO.
You can use cifmw_extras the same way as it is used in other jobs -- you will need to add a task that includes those vars from files.

ci/run_playbooks.yml
```diff
@@ -5,15 +5,21 @@
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
-  vars:
-    fvt_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
-    ci_framework_logs: "{{ ansible_env.HOME }}/ci-framework-data/logs"
+  vars_files:
```
elfiesmelfie, line 8: vars_files is new to me. I need to go check it out

mgirgisf:
recheck

mgirgisf:
recheck

ci/vars/common.yml
```diff
@@ -0,0 +1,3 @@
+fvt_dir: "{{ ansible_env.HOME }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}"
+logs_dir: "{{ ansible_env.HOME }}/ci-framework-data/logs/functional-tests"
+env_vars: "/tmp/env_vars.yml"
```
elfiesmelfie, line 3: can you name this "env_vars_file" so it's clearer what it does?

elfiesmelfie (changes requested):

ci/ansible.cfg
elfiesmelfie, line 1: The change to this file could go in on its own, since it effects this changes as well: https://github.com/infrawatch/feature-verification-tests/pull/90/files#diff-94e99cdfa193887306a39d87dd8f1eff431b6dbc42547fe3ff0ced3405e699d5

ci/report_result.yml
elfiesmelfie, line 1: It looks like this file isn't used.

Where is the pass/fail  determined now?

i.e. Does the scope of this PR still cover marking the job as passed/failed dependent on the contents of test_result_out?

ci/run_playbooks.yml
```diff
@@ -56,18 +62,20 @@
...
             - setup
 
-    - name: "Run a playbook locally on the target host"
-      ansible.builtin.shell:
-        cmd: |
-          ANSIBLE_CONFIG=ci/ansible.cfg ansible-playbook -v -e "cifmw_openshift_kubeconfig={{ cifmw_openshift_kubeconfig }}" -e "cifmw_path={{ cifmw_path }}" ci/telemetry_autoscaling.yml
-        chdir: "{{ fvt_dir }}"
-      register: output
+    - name: Run telemetry autoscaling test
+      block:
```
elfiesmelfie, line 66: What is the reason for adding this to a block?
I don't see an obvious reason for it.

elfiesmelfie (changes requested):

ci/run_playbooks.yml
mgirgisf: Just for Grouping the Related Tasks for running the autoscaling playbook.

ci/report_result.yml
mgirgisf: the use of this file is still commented until we merge the callback_plugin changes.

elfiesmelfie (approved):

PR #89, opened 2024-05-22
[custom_logger] Update the testID search

elfiesmelfie:


elfiesmelfie:
To test the updates, you can use this playbook:

```
$ cat test_logger.yml 
---
- hosts: localhost
  tasks:
    - name: RHOSO-1 show some message
      debug: msg="some"

    - name: Show a var, this is the RHOSO-2 test
      debug: msg="some_var"

    - name: |
        Check how a multiline string task name is
        handles for RHOSO-3
      debug: msg="foo"

    - name: try a test where there is no ID, just the RHOSO prefix
      debug: msg="bar"

    - name: Need to check for RHELOSP-42 test ID
      debug: msg="eventually, we want to look for one prefix only, but it should be configurable"

    - name: What happens when the ID is at the end RHELOSP-43
      debug: msg="Oh! this actually does work"

```


To run the playbook, use 
```
$ cd ci/ &&  rm -rf test_run_result.out && ansible-playbook test_logger.yml -v  && cat test_run_result.out
```

elfiesmelfie:
recheck

elfiesmelfie:
The test playbook is running here: https://github.com/infrawatch/feature-verification-tests/pull/90

PR #93, opened 2024-06-03
Add zuul report

mgirgisf:
This PR gurante that the zuul job will fail in case of any failure happened in the tasks with the test cases IDs
https://github.com/infrawatch/feature-verification-tests/pull/91

And it generate 2 log files that is being uploaded to the zuul logs.:
1- summary_results.log  that contains a summary report : ci-framework-data/logs/functional-tests/summary_results.log
2-test_run_result.log that has the old format for each task and its state.  ci-framework-data/logs/functional-tests/test_run_result.log

PR #99, opened 2024-06-11
Add logging tests playbooks.

ayefimov-1:
Added logging test playbooks


ci/telemetry_logging.yml
mgirgisf, line 1: I would suggest to add import_role directly to telemetry_autoscaling.yml (I will create a PR to rename it to osp18-functional-tests.yml) as if you checked the logs you can see that the hook is not triggering this playbook.

.zuul.yaml
mgirgisf, line 1: No need to create a new job, or you need to have a separate hook and alot of duplications its better to add the include role to the existing job

mgirgisf (changes requested):

ci/telemetry_logging.yml
elfiesmelfie: @mgirgisf Thanks for proposing a rename of the playbook we use.
https://github.com/infrawatch/feature-verification-tests/pull/100

@ayefimov-1 once PR#100 is merged, you can simplify this PR to remove the zuul updates.
You won't need to define a new zuul job anymore, since the autoscaling and logging roles can be run in a single job.

ci/telemetry_logging.yml
```diff
@@ -0,0 +1,33 @@
+---
```
elfiesmelfie, line 1: FYI PR#100 will render this file un-needed

roles/telemetry_logging/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Requirements
+------------
+The following resources are required in the OpenStack cloud
+* network called public
+* network called provate
+* security group called basic with ssh and tcp enabled to the VMs
+* a flavor called m1.small
+* an image called cirros
+
```
elfiesmelfie, line 14: These are not required for the logging role as far as I know.

They are required for the telemetry_autoscaling role because the heat templates need these resources to exist.

roles/telemetry_logging/tasks/hosts_logging.inv
elfiesmelfie, line 1: This file should be removed.

For running in CI, zuul provides an inventory file.
For running locally, the values configured will vary, and should be provided by the user

roles/telemetry_logging/tasks/logging_endpoint_tests.yml
```diff
@@ -0,0 +1,24 @@
...
+- name: Get Endpoint
+  ansible.builtin.shell:
+    cmd: |
+      oc project openstack     
+      kubectl exec openstackclient -- openstack endpoint list --service="{{ item[0] }}" --service="{{ item[1] }}"  --interface="{{ item[2] }}"
+  register: result 
+  changed_when: false
+  ignore_errors: yes
+
+- name: Set status "{{ item }}"
```
elfiesmelfie, line 12: This set of tasks is repeated across multiple files.
it is not necessary IMHO, since we have updates to the custom logger that will provide some reporting.

Please take a look at https://github.com/infrawatch/feature-verification-tests/pull/93 to see what output the custom_logger plugin produces. I recommend making adjustments there which will run to produce reports on the tests.

roles/telemetry_logging/tasks/logging_journal_tests.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+- name: Get journals
+  ansible.builtin.shell:
+    cmd:
+      tstamp=$(date -d '30 minute ago' "+%Y-%m-%d %H:%M:%S") 
+      journalctl -t "{{ item }}" --no-pager -S "${tstamp}" | wc -l
+  register: journal_wc
+  changed_when: false
+  failed_when: false
```
elfiesmelfie, line 10: This pattern of using failed_when: false should be revised.

The roles in this repo are typically included by playbooks using ``ignore_errors: true``, which means that all the tasks will run, regardless of pass/fail results.

I would recommend adding the fail condition to the test task directly, instead of using a later task to check the status.

I see that you're trying to add reporting, however, we have the existing functionality to do this without adding additional tasks for each test.
The custom logger was implemented to resolve a similar issue, where we wanted to record the pass/fail status for certain tasks into a file  in order to report to polarian. Recently, Marihan has added some additional functionality to produce a test report that isn't limited to tagged tasks.

roles/telemetry_logging/tasks/logging_pod_tests.yml
```diff
@@ -0,0 +1,32 @@
...
+  ansible.builtin.set_fact:
+     exists: FAILED
+
+- name: Test pod exists - "{{ item }}"
+  ansible.builtin.set_fact:
+    exists: PASSED
+  when: 
+    - output.rc == 0
+    - podinstance.stdout != ""
+
```
elfiesmelfie, line 28: I have a suggestion for improving the efficiency here.

if you use the ``failed_when`` option for the "Get pod instance" task, you can then set the var ``exists`` based on the status of that test.

The success filter should be useful here.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/success_test.html
you can do something like:

```
- set_fact:
     exists: "{{ output is ansible.builtin.success }}"
```

roles/telemetry_logging/tasks/logging_pod_tests.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Push results to "{{ pod_out_file }}" 
+  ansible.builtin.shell:
+    'echo "{{ item }},{{ podinstance.stdout }},{{ exists }}" >> "{{ pod_out_file }}"'
```
elfiesmelfie, line 31: You  should look into a few of the other modules in ansible instead of using shell all the time. 
ansible lint can sometimes be very useful here, since some of the rules include these kind of suggestions.

.zuul.yaml
ayefimov-1: Removed job and updated job descriptor of the autoscaling job.

roles/telemetry_logging/tasks/main.yml
```diff
@@ -0,0 +1,500 @@
...
+    - name: "Log into OCP"
+      ansible.builtin.shell:
+        cmd: |
+          oc login -u kubeadmin -p "12345678" https://api.crc.testing:6443
+      failed_when: false
+      changed_when: false
+
+
+- name: 'Verify projects exist'
+  hosts: localhost
```
elfiesmelfie, line 23: So, this main.yml should not be a series of plays, and it should not reference hosts.

However, the role is going to be imported in a playbook somewhere, and the hosts, etc will have to be configured.

Typically, I think "controller" is used for localhost.

When running on Zuul, `localhost` refers to the zuul executer node, which we don't want to run stuff on in general, since it's not part of the test environment. 

There are some other nodes listed in this file as well, and I'll add some notes to the relevant lines below

roles/telemetry_logging/tasks/main.yml
```diff
@@ -0,0 +1,500 @@
...
+        state: absent
+      changed_when: false
+
+    - name: Run packagemanifest tests
+      ansible.builtin.include_tasks: logging_subscription_tests.yml
+      loop: "{{ subscription_list }}"
+
+
+- name: Verify ctlplane node containers
+  hosts: ctlplane_nodes
```
elfiesmelfie, line 410: EDPM nodes are in the `computes` group, with individual hosts called `compute-0`, `compute-1`, etc

Zuul compiles an inventory file and logs it for convenience:
https://logserver.rdoproject.org/99/99/d89a0b5d52bc318c9ff9d350160d251647101f33/github-check/functional-logging-tests-osp18/37131c1/zuul-info/inventory.yaml

```suggestion
  hosts: ctlplane_nodes
```
```suggestion
  hosts: ctlplane_nodes
```
```suggestion
  hosts: computes
```

roles/telemetry_logging/tasks/main.yml
```diff
@@ -0,0 +1,500 @@
...
+    - name: Copy output file to hypervisor
+      ansible.builtin.fetch:
+        src: "/root/{{ container_out_file }}"
+        dest: ./
+        flat: true
+      changed_when: false
+
+
+- name: Verify ctlplane node files
+  hosts: ctlplane_nodes
```
elfiesmelfie, line 439: Same as above, should be `computes`

elfiesmelfie (changes requested):
Check out the PRs I listed, and update the format of main.yml, the docs that go with the role, and the zuul configs

elfiesmelfie:
Here's an example of the Zuul reports that come with https://github.com/infrawatch/feature-verification-tests/pull/93: https://logserver.rdoproject.org/93/93/cdefbb04e71f6878b47d8b41cec6546f13ca7a6c/github-check/functional-tests-on-osp18/8ed4d7e/controller/ci-framework-data/logs/functional-tests/

roles/telemetry_logging/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+# defaults file for telemetry_logging
```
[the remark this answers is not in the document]
ayefimov-1: Done

roles/telemetry_logging/tasks/logging_endpoint_tests.yml
ayefimov-1: Done

ci/telemetry_logging.yml
ayefimov-1: removing file.

roles/telemetry_logging/tasks/main.yml
```diff
@@ -0,0 +1,500 @@
...
+        state: absent
+      changed_when: false  
+
+    - name: Verify projects created
+      ansible.builtin.include_tasks: logging_proj_tests.yml
+      loop: "{{ proj_list }}"
+
+
+- name: Verify Openstack endpoints
+  hosts: localhost
```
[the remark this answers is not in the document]
elfiesmelfie: I am mistaken here. We may use ``localhost`` for these because of the unconventional way we execute the tests. i.e. we use a pattern of:

`- command: ansible-playbook ...`

However, the information about the zuul executer, the hostnames, etc are still accurate

roles/telemetry_logging/tasks/main.yml
```diff
@@ -0,0 +1,500 @@
...
+        state: absent
+      changed_when: false
+
+    - name: Verify journalctl identifiers 
+      ansible.builtin.include_tasks: logging_journal_tests.yml
+      loop: "{{ identifiers_list }}"
+
+    - name: Copy output file to hypervisor
+      ansible.builtin.fetch:
+        src: "/root/{{ journal_exists_outfile }}"
```
elfiesmelfie, line 487: This file path will need to be changed.

I assume it's in /root because you're using `become: true` and the file ends up in the home directory.

You can replace `/root...` with `"{{ ansible_env.HOME }}/{{ journal_exists_outfile }}"`

However, in general, it would be more useful to define a log_output_dir, so you can configure that directory more easily.

I think that this outputfile  may become obsolete once we switch to using the custom_logger for running the telemetry_logging role.

PR #102, opened 2024-06-14
[zuul] Use alternative syntax for defining hooks

elfiesmelfie:
ci-framework allows for hooks to be defined as multiple hooks in a list, or single hooks in their own parameters.
If two separate vars files (passed to cifmw_extras) define the same hook, then one gets overwritten by the other, depending on the order in which the vars files are imported.

Using the "single hook in its own parameter" method of defining the hook prevents this issue.


mgirgisf (approved):
Thanks Emma

PR #106, opened 2024-06-21
Add logging test playbooks/roles to master

ayefimov-1:
created telemetry_logging and qe_common roles to enable logging tests in zuul.


ci/telemetry_logging.yml
```diff
@@ -0,0 +1,33 @@
...
+    - vars/common.yml
+    - vars/osp18_env.yml
+  tasks:
+    - name: "Run logging Tests"
+      ansible.builtin.import_role:
+        name: telemetry_logging
+
+        - import_role:
+            name: telemetry_logging
+            tasks_from: main
```
elfiesmelfie, line 18: The default value is "main"

I would suggest refactoring the main.yml to have two sections.
For all compute-node tests, use "when: telemetry_logging_run_compute_tests | bool"
For all controller tests, use "when: telemetry_logging_run_controller_tests | bool"

This means you can have separate plays in this playbook, which import the same role, but pass different vars.

e.g. there would be multiple plays in this playbook.

```
- name: run controller tests
  hosts: controller_nodes # please change to the right value
  tasks:
    - import_role:
          name: telemetry_logging
          vars:
                telemetry_logging_run_compute_tests: true

- name: Run compute tests
  hosts: compute_nodes ## check the name
  tasks:
      - import_role:
          name: telemetry_logging
          vars:
              telemetry_logging_run_controller_tests: true
```

The info in this suggestion is incomplete, you will need to test it, and add the appropriate missing information

ci/telemetry_logging.yml
```diff
@@ -0,0 +1,33 @@
...
+    PATH: "{{ cifmw_path }}"
+  vars_files:
+    - vars/common.yml
+    - vars/osp18_env.yml
+  tasks:
+    - name: "Run logging Tests"
+      ansible.builtin.import_role:
+        name: telemetry_logging
+
+        - import_role:
```
elfiesmelfie, line 16: This task in not well formatted, and needs to be removed

logging_tests_ctlplane.yml
```diff
@@ -0,0 +1,108 @@
+- name: Prepare Logging Tests
```
elfiesmelfie, line 1: I'm not sure where this is called from yet

roles/qe_common/README.md
```diff
@@ -0,0 +1,42 @@
+telemetry_logging
```
elfiesmelfie, line 1: This needs to be updated to match the right role information

roles/qe_common/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
elfiesmelfie, line 1: These are from a different role, they should be removed.

.zuul.yaml
```diff
@@ -32,10 +32,46 @@
...
+
+- job:
+    name: functional-logging-tests-osp18
+    parent: telemetry-operator-multinode-logging
+    description: |
+      Run the logging functional test on osp18+patched version
+    vars:
+      cifmw_extras:
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir  }}/scenarios/centos-9/multinode-ci.yml"
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-autoscaling.yml"
```
elfiesmelfie, line 44: This should be removed, we don't need the autoscaling vars for logging.

ayefimov-1: ack.

ci/telemetry_logging.yml
ayefimov-1: replaces with run_playbooks_logging.yml

ci/telemetry_logging.yml
ayefimov-1: ack

logging_hosts.inv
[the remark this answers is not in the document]
ayefimov-1: ack

ayefimov-1:
closing merge request.

PR #119, opened 2024-07-04
Update the command for oc patch

mgirgisf:


ci/run_playbooks.yml
```diff
@@ -39,7 +39,7 @@
         - name: Patch the openstackversions to use the master containers for aodh and heat
           ansible.builtin.shell:
             cmd: |
-              oc patch openstackversions openstack-galera-network-isolation --type merge --patch-file ci/patch-openstack-versions.yaml
+              oc patch  oscp/controlplane  --type merge --patch-file ci/patch-openstack-versions.yaml
```
elfiesmelfie, line 42: This is originally a patch to the openstackversions CR, not to the OpenStackControlePlane.

Has openstackversions been removed?
It probably has been renamed, which should be addressed, but please doublecheck the resource type

elfiesmelfie (changes requested):

ci/run_playbooks.yml
elfiesmelfie, line 42:
```suggestion
              oc patch  openstackversions controlplane  --type merge --patch-file ci/patch-openstack-versions.yaml
```

roles/telemetry_autoscaling/tasks/post_teardown.yml
```diff
@@ -27,4 +27,4 @@
 - name: Revert the version update
   ansible.builtin.shell:
     cmd: |
-      oc patch openstackversions openstack-galera-network-isolation --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
\ No newline at end of file
+      oc patch  oscp/controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
```
elfiesmelfie, line 30:
```suggestion
      oc patch  openstackversions controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
```

elfiesmelfie (changes requested):

ci/run_playbooks.yml
mgirgisf: my mistake thanks emma I updated it

mgirgisf:
recheck

mgirgisf:
recheck

mgirgisf:
@elfiesmelfie can I have your feedback on that so we can merge it, it also effect the downstream Job https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/tripleo-ci-internal/build/679cb8dbf27043399bca3b231d5e3707/logs

ci/run_playbooks.yml
```diff
@@ -81,6 +81,13 @@
             content: "{{ output.stdout }}"
             dest: "{{ logs_dir }}/ansible_run.log"
 
+
+    - name: Revert the version update
+      ansible.builtin.shell:
+        cmd: |
+          oc patch  openstackversions controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
+      when: "{{ patch_openstackversions | bool }}"
```
elfiesmelfie, line 89: This looks reasonable.
My only concern is that if some tasks fail, that the patch won't be reverted.

It might be worth adding a handler or a try/except block to make sure this is always run.
A handler would be cleaner, but require some more testing, I think. 
You'd have to attach a handler to the patch task, and then it would be run at the end of the play, I think.
This would be a good follow-up learning task, and I'm not going to block merging this based on not having a handler.

elfiesmelfie (approved):

ci/run_playbooks.yml
mgirgisf: I will create a jira ticket for that

PR #132, opened 2024-07-22
Add a custom_junit callback plugin

mgirgisf:
This custom_junit plugin extends the existing junit plugin and produces an XML report with classes and names that polarian expects.

### DIFF
--- a/callback_plugins/custom_junit.py
+++ b/callback_plugins/custom_junit.py
@@ -17,12 +17,9 @@ class CallbackModule(JunitCallbackModule):
         super(CallbackModule, self).__init__()
 
         # Custom environment variable handling
-        # self._output_dir = os.getcwd()
         # Update this to parse these values from the config file, as well as the env.
         self._output_dir = os.path.expanduser("~/.ansible.log")
-        # self._test_case_prefix = os.getenv('JUNIT_TEST_CASE_PREFIX', '[TEST]')
         self._test_case_prefix = os.getenv('JUNIT_TEST_CASE_PREFIX', 'TEST')
-        #self._fail_on_ignore = os.getenv('JUNIT_FAIL_ON_IGNORE', 'False').lower()
         self._fail_on_ignore = 'true'  # this is needed because we use "ignore_errors" on the playbooks so that all the tests are run
         self._include_setup_tasks_in_report = os.getenv('JUNIT_INCLUDE_SETUP_TASKS_IN_REPORT', 'False').lower()
         self._hide_task_arguments = os.getenv('JUNIT_HIDE_TASK_ARGUMENTS', 'True').lower()
@@ -37,7 +34,6 @@ class CallbackModule(JunitCallbackModule):
     def _finish_task(self, status, result):
         """ record the results of a task for a single host """
         task_uuid = result._task._uuid
-
         if hasattr(result, '_host'):
             host_uuid = result._host._uuid
             host_name = result._host.name
@@ -62,7 +58,7 @@ class CallbackModule(JunitCallbackModule):
         if self._test_case_prefix in task_data.name:
             task_data.add_host(HostData(host_uuid, host_name, status, result))
 
-        # This bit is new
+        # Debugging
         if task_data.name.startswith(self._test_case_prefix):
             print(f"This task ({task_data.name}) starts with the test_prefix({self._test_case_prefix})")
         if self._test_case_prefix in task_data.name:
@@ -71,37 +67,23 @@ class CallbackModule(JunitCallbackModule):
             print(f"This task ({task_data.name}) failed, but may not be reported")
 
     def mutate_task_name(self, task_name):
-        #print("enter mutate_task_name(task_name=%s)" % task_name)
-
+        # Debugging
         if not self._test_case_prefix in task_name:
             print("task_name (%s) does not contain prefix (%s)" % (task_name, self._test_case_prefix))
-
         new_name = task_name
         new_name = new_name.split("\n")[0]  # only use the first line, so we can include IDs and additional description
-        #print("%s\t(take the first line of the task name)" % new_name)
-
         # this covers when a task is included, but the including task is the one that is the test
         new_name = new_name.split(":")[-1]  # only provide the last part of the name when the role name is included
-        #print("%s\t(split at :, take last element)" % new_name)
 
         if len(self._test_case_prefix) > 0:
             # this one may not be needed...
             new_name = new_name.split(self._test_case_prefix)[-1]  # remove the test prefix and everything before it
-            #print("%s\t(remove test prefix)" % new_name)
 
         new_name = new_name.lower()
-        #print("%s\t(lowercase)" % new_name)
-
         new_name = re.sub(r'\W', ' ', new_name)  # replace all non-alphanumeric characters (except _) with a space
-        #print("%s (Replace non-alphanumerics with a space)" % new_name)
-
         new_name = re.sub(r'(^\W*|\W*$)', '', new_name)  # trim any trailing or leading non-alphanumeric characters
-        #print("%s\t(trim leading or trailing characters" % new_name)
-
         new_name = re.sub(r' +', '_', new_name)  # replace any number of spaces with _
-        #print("%s\t(spaces -> _)" % new_name)
 
-        #print("exit mutate_task_name")
         return new_name
 
     def _build_test_case(self, task_data, host_data):
@@ -109,7 +91,6 @@ class CallbackModule(JunitCallbackModule):
            This is used in generate_report. The task_data and host data will get passed.
         """
         # Use the original task name to define the final name
-        #new_name = self.mutate_task_name(task_data.name)
 
         print("%s\t(task_name, pre-_build_test_case)" % task_data.name)
         tc = super()._build_test_case(task_data, host_data)
@@ -118,11 +99,10 @@ class CallbackModule(JunitCallbackModule):
 
         print("%s\t(tc.name, post-mutate_task_name)" % tc.name)
 
-        # I don't want these properties for now; I may be able to omit them with a config option
+        # These can be able to omit with a config option
         # These two control whether testcases contain the system_out and
         # system_err elements that show STDOUT and STDERR
         tc.system_out = None
         tc.system_err = None
-
         tc.classname = "openstack-observability"
         return tc
--- a/callback_plugins/custom_junit.py
+++ b/callback_plugins/custom_junit.py
@@ -1,0 +1,128 @@
+from ansible.plugins.callback.junit import CallbackModule as JunitCallbackModule
+from ansible.plugins.callback.junit import HostData
+
+import os
+import time
+import re
+from ansible.utils._junit_xml import TestCase, TestError, TestFailure, TestSuite, TestSuites
+
+
+class CallbackModule(JunitCallbackModule):
+    """
+    Custom callback that overrides the default JUnit callback
+    """
+    CALLBACK_NAME = 'custom_junit'
+
+    def __init__(self):
+        super(CallbackModule, self).__init__()
+
+        # Custom environment variable handling
+        # self._output_dir = os.getcwd()
+        # Update this to parse these values from the config file, as well as the env.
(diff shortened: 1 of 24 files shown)


mgirgisf:
recheck

roles/telemetry_autoscaling/tasks/configure_heat.yml
```diff
@@ -27,19 +27,48 @@
...
+      lineinfile:
+        path: "{{ logs_dir }}/test_run_result.xml"
+        line: '<testcase classname="autoscaling" name="RHOSO-12657 config heat template" />'
+        insertbefore: "</testsuite>"
+
+  rescue:
+    - name: Add failure to test result
+      delegate_to: localhost
+      lineinfile:
+        path: "{{ logs_dir }}/test_run_result.xml"
```
elfiesmelfie, line 61: The buildset logs are showing that this doesn't exist

The logs at https://logserver.rdoproject.org/32/132/8c02c82c6426d08d090b9ae8a036610ab59f5c6f/github-check/functional-tests-on-osp18/3ccb604/controller/ci-framework-data/logs/functional-tests/
have an autoscaling_results file.
Update this to match.

roles/telemetry_autoscaling/tasks/configure_heat.yml
```diff
@@ -27,19 +27,48 @@
...
+        insertbefore: "</testsuite>"
+
+    - name: increment number of failures
+      set_fact:
+        number_of_failures: "{{ number_of_failures | int + 1 }}"
+
+- name: increment number of tests
+  set_fact:
+    number_of_tests: "{{ number_of_tests | int + 1 }}"
+
```
elfiesmelfie, line 74: Can these be done in v2_playbook_on_stats?

callback_plugins/custom_logger.py
```diff
@@ -1,92 +1,274 @@
+# (c) 2016 Matt Clay <matt@mystile.com>
+# (c) 2017 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
elfiesmelfie, line 3: Copy/pasting the existing junitXML plugin is not the way to go.

Either try to contribute upstream to it, or inherit from the existing class here to make changes.

There MIGHT be an issue with using a GPL-licensed library to provide the base class, since this project is not GPL licensed, nor do we plan to adopt GPL v3 as the license.

elfiesmelfie: Yet another alternative would be to record the results and then render the XML at the end using a jinja2 template, so that there's no back-tracking through the document. However, this would use more memory.

roles/telemetry_autoscaling/tasks/configure_heat.yml
```diff
@@ -27,19 +27,48 @@
...
+      ansible.builtin.template:
+        dest: ~/templates/autoscaling/vnf/template.yaml
+        src: template.yaml.j2
+
+    - name: Add pass to test results
+      delegate_to: localhost
+      lineinfile:
+        path: "{{ logs_dir }}/test_run_result.xml"
+        line: '<testcase classname="openstack-observability"  name="configure-hest-template-automatically-scaling-instances"  time="1"/>'
+        insertbefore: "</testsuite>"
```
elfiesmelfie, line 55: This is the only test result file I can see.

I need to look for the one generated by the junit plugin

callback_plugins/custom_logger.py
```diff
@@ -1,92 +1,274 @@
...
-                f.write(f"  - {task_name}\n")
-            f.write("Succeeded Tasks:\n")
-            for task_name in self.results[host]['ok_task_names']:
-                f.write(f"  - {task_name}\n")
 
-    def v2_runner_on_ok(self, result):
-        host = result._host.get_name()
-        task_name = result._task.get_name()
-        self.log_task_result(host, 'passed', task_name)
+        self._output_dir =  os.getcwd()
```
elfiesmelfie, line 53: This is not a sensible default. I don't know where this ends up, maybe on the executer, but it's not in a location that gets uploaded

elfiesmelfie (commented):
I ran this locally, using the test playbook for the custom logger

I left ci/ansible.cfg as-is, and used the following command to test`
cd ci && ansible-playbook -v github/test_logger.yml`

The output was this:
```
cat *.xml
<?xml version="1.0" ?>
<testsuites disabled="0" errors="0" failures="0" tests="1" time="0.012203693389892578125">
	<testsuite disabled="0" errors="0" failures="0" name="test_logger" skipped="0" tests="1" time="0.012203693389892578125">
		<testcase classname="/home/emma/repos/github/infrawatch/feature-verification-tests/ci/github/test_logger.yml:6" name="[localhost] Run tests to make sure the custom_logger plugin is running as expected: RHOSO-01 show some message" time="0.012203693389892578">
			<system-out>{
    &quot;changed&quot;: false,
    &quot;msg&quot;: &quot;You can start the task name with the test ID.&quot;
}</system-out>
		</testcase>
	</testsuite>
</testsuites>
```

It looks like unexpected values are being used in unexpected places, and the TestCase class probably needs to be updated.
There's also only one testcase recorded because there's only one task with the ID at the start of the name.

callback_plugins/custom_logger.py
```diff
@@ -1,92 +1,274 @@
...
+        self._task_class = os.getenv('JUNIT_TASK_CLASS', 'False').lower()
+        self._task_relative_path = os.getenv('JUNIT_TASK_RELATIVE_PATH', '')
+        self._fail_on_change = os.getenv('JUNIT_FAIL_ON_CHANGE', 'False').lower()
+        self._fail_on_ignore = os.getenv('JUNIT_FAIL_ON_IGNORE', 'False').lower()
```
elfiesmelfie, line 57: This should be true, because we use "ignore_errors: true" in our playbooks

elfiesmelfie (commented):
We can extend the existing xml plugin, and redefine only what we need.
It should be possible to create a new class, with the existing XML plugin as a base and then redefines the methods we need to update.

callback_plugins/custom_junit.py
```diff
@@ -0,0 +1,49 @@
+from ansible.plugins.callback import CallbackModule as OriginalCallbackModule
```
elfiesmelfie, line 1: Use ansible.plugins.callback.junit.CallbackModule as the superclass

callback_plugins/custom_junit.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+import time
+import re
+from ansible import constants as C
+from ansible.module_utils._text import to_bytes, to_text
+from ansible.utils._junit_xml import TestCase, TestError, TestFailure, TestSuite, TestSuites
+
+class CustomCallbackModule(OriginalCallbackModule):
```
elfiesmelfie, line 9: CallbackModule is a required classname for ansible to run the module.

elfiesmelfie:
recheck

elfiesmelfie:
I can see that the junit plugin actually ran, the logs are in: https://logserver.rdoproject.org/32/132/496cae2c9347782c5596867f8e0097162186d67a/github-check/functional-tests-on-osp18/39d2e28/controller/ci-framework-data/logs/ci_script_011_run_run_functional.log

HOWEVER, there is no XML plugin generated anywhere under ~ (there is only the sample one that was added to this PR)

ci/run_autoscaling_osp18.yml
```diff
@@ -60,14 +61,49 @@
...
+            dest: "{{ logs_dir }}/autoscaling/ansible_run.log"
+
+      always:
         - name: Revert the version update
           ansible.builtin.shell:
             cmd: |
               oc patch  openstackversions controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
           when: "{{ patch_openstackversions | bool }}"
 
+        - name: "Where is that XML file!?"
```
elfiesmelfie, line 87: I _JUST_ realised why this isn't working. 

The junit callback create the output file only at the end of the playbook execution.

We need to run the report result as a a separate hook AFTER  the other tests are run.

elfiesmelfie: Technically, report result can stay where it is. We just need to have a post_test hook that runs to do the copying of the data.
This should be temporary, since the output_dir can be configured.

elfiesmelfie:
The xml file is not as expected: https://logserver.rdoproject.org/32/132/4037b06f8a448cf65c58a192c3db5c3079a56ad7/github-check/functional-tests-on-osp18/7494a4e/controller/ci-framework-data/logs/functional-tests/ansible-tests/controller-run_autoscaling_osp18-1725481538.823415.xml

There should be more tests in there

elfiesmelfie:
recheck

vyzigold:
recheck - The fix in heat-operator merged, so the autoscaling test should work.

elfiesmelfie:
recheck

tosky (commented):
Two quick questions:
- there are >40 commits in this review, are all needed or can the history be cleaned a bit?
-  the internal ID for each test (RHOSO-...) are not needed and should probably be removed.

elfiesmelfie:
> Two quick questions:
> 
>     * there are >40 commits in this review, are all needed or can the history be cleaned a bit?
> 
>     * the internal ID for each test (RHOSO-...) are not needed and should probably be removed.

I'll squash and merge; we typically do not rebase the PR branches, so that we keep the comment history as much as possible. The changes then get squashed and merged.
The test IDs are needed to get a pass/fail at the moment. There's a second logging plugin that logs the tests that have those IDs and creates a test_run_result.out file with $TESTID=[passed|failed]. the report_result.yml playbooks is then run to check for failures.
We need to decide what to do with them, but it is not part of the scope of this change to do so.

callback_plugins/custom_junit.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self):
+        super(CallbackModule, self).__init__()
+
+        # Custom environment variable handling
+        # self._output_dir = os.getcwd()
+        # Update this to parse these values from the config file, as well as the env.
+        self._output_dir = os.path.expanduser("~/.ansible.log")
+        # self._test_case_prefix = os.getenv('JUNIT_TEST_CASE_PREFIX', '[TEST]')
+        self._test_case_prefix = os.getenv('JUNIT_TEST_CASE_PREFIX', 'TEST')
+        #self._fail_on_ignore = os.getenv('JUNIT_FAIL_ON_IGNORE', 'False').lower()
```
vyzigold, line 25: The commented lines of code above here seem like something maybe isn't finished. Are they needed? Is it some debugging leftover? I think we should delete them if there isn't a reason for them to stay.

callback_plugins/custom_junit.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # This bit is new
+        if task_data.name.startswith(self._test_case_prefix):
+            print(f"This task ({task_data.name}) starts with the test_prefix({self._test_case_prefix})")
+        if self._test_case_prefix in task_data.name:
+            print(f"This task ({task_data.name}) should be reported because it contains test_prefix({self._test_case_prefix})")
+        if status == 'failed':
+            print(f"This task ({task_data.name}) failed, but may not be reported")
+
+    def mutate_task_name(self, task_name):
```
vyzigold, line 73: same thing here with all the commented code throughout this whole function

vyzigold (dismissed):

elfiesmelfie:
This looks fine as long as https://github.com/infrawatch/feature-verification-tests/pull/159  passes.
https://github.com/infrawatch/feature-verification-tests/pull/159 is based on this branch and adds the "TEST" prefix to some of the tasks

mgirgisf:
> This looks fine as long as #159 passes. #159 is based on this branch and adds the "TEST" prefix to some of the tasks

I already tested also with previous commit " https://github.com/infrawatch/feature-verification-tests/compare/22a919ac53acc542d9bc78cceb79302af53c660c..d96a6904000d5ace5bbefb926bd914420da50b48"

elfiesmelfie (approved):

PR #134, opened 2024-07-29
[zuul] Add graphing Job

mgirgisf:


mgirgisf:
recheck

roles/telemetry_graphing/handlers/main.yml
elfiesmelfie, line 1: You can remove the handlers dir if it's not used

roles/telemetry_graphing/README.md
```diff
@@ -0,0 +1,38 @@
...
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+    - hosts: servers
+      roles:
+         - { role: username.rolename, x: 42 }
+
+License
+-------
+
+BSD
```
elfiesmelfie, line 33: Should be apache 2

roles/telemetry_graphing/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
+=========
+
+A brief description of the role goes here.
```
elfiesmelfie, line 4: Please update this README to add a description of the role, the requirements, vars, etc

roles/telemetry_graphing/vars/main.yml
elfiesmelfie, line 1: Remove this if it's not used

ci/run_graphing_test.yml
```diff
@@ -0,0 +1,16 @@
...
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  vars_files:
+    - vars/common.yml
+  tasks:
+    - name: "Run Graphing Console UI tests"
+      ansible.builtin.import_role:
+        name: telemetry_graphing
+
```
elfiesmelfie, line 14: The typical approach is to use "ignore_errors: true" so that all the tests get run, then the report_result will check for failures.

roles/telemetry_graphing/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: "Display VM creation results"
+  ansible.builtin.debug:
+    msg: "VM creation result: {{ item.stdout }}"
+  loop: "{{ vm_creation_result.results }}"
+
+- name: "[Test] verify url for graphing test"
+  ansible.builtin.include_tasks: 
+    file: verify_url.yml
+
+- name: "[Test] verify promotheus query"
```
elfiesmelfie, line 26: I think that this will always pass as long as the file is successfully included.

elfiesmelfie: So the test should be in one of the tasks in the included file

elfiesmelfie (changes requested):

PR #136, opened 2024-08-02
[WIP] Add  jUNitXML plugin to create the output xml file

myadla:
Going forward polarion supports only xml reporting. Hence testing jUNitXML plugin in ansible to o/p logs into xml format

OSPRH-8374

### DIFF
--- a/ci/ansible.cfg
+++ b/ci/ansible.cfg
@@ -1,5 +1,5 @@
 [defaults]
-callbacks_enabled = custom_logger
+callbacks_enabled = custom_logger, junit
 callback_plugins = ../callback_plugins
 # additional paths to search for roles
 roles_path = ../roles:/usr/share/ansible/roles:/etc/ansible/roles:~/.ansible/roles:../../ci-framework/roles
--- a/ci/osp18_functional_tests.yml
+++ b/ci/osp18_functional_tests.yml
@@ -5,6 +5,7 @@
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
+    JUNIT_TEST_CASE_PREFIX: "testsuites"
   tasks:
     - name: "Run Telemetry Autoscaling tests"
       ansible.builtin.import_role:
--- a/ci/report_result.yml
+++ b/ci/report_result.yml
@@ -4,6 +4,7 @@
     cmd: |
       cp  {{ fvt_dir }}/test_run_result.out {{ logs_dir }}/test_run_result.log
       cp  {{ fvt_dir }}/summary_results.log {{ logs_dir }}/summary_results.log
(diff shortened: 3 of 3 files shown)


elfiesmelfie (changes requested):
For setting the env vars, you can add `environment` to a play, or you can add environment to a task that uses shell. 

This would be an easier implementation, would limit the scope of the env update to a single play or task, rather than making permanent changes to the system by using /etc/environment., and would be easier to understand and maintain


https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_environment.html

roles/telemetry_autoscaling/tasks/set_env_var.yml
```diff
@@ -0,0 +1,16 @@
+---
+# Set environment variables for JUNIT callback plugin.
+# It writes playbook output to a JUnit formatted XML file.
+- name: envt vars
+  set_fact:
+    os_environment:
+      - key: JUNIT_TEST_CASE_PREFIX
```
mgirgisf, line 7: we don't need to add it to testsuites , it will make it filter only tasks start with that

ci/ansible.cfg
```diff
@@ -1,5 +1,5 @@
 [defaults]
-callbacks_enabled = custom_logger
+callbacks_enabled = custom_logger, junit
```
elfiesmelfie, line 2: How does this relate to the change here: https://github.com/infrawatch/feature-verification-tests/pull/132
Are they alternatives, or complimentary?

mgirgisf: its not related, in this PR Munesha is enable the callback plugin only, but in my PR I was propose another solution  trying to customise the plugin.  as I have the feeling that we can't use the default one directly .

ci/osp18_functional_tests.yml
```diff
@@ -5,6 +5,7 @@
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
+    JUNIT_TEST_CASE_PREFIX: "testsuites"
```
elfiesmelfie, line 8: Can we use "[test]" instead?

myadla: @elfiesmelfie Yes we can .. 
but defining "JUNIT_TEST_CASE_PREFIX" here is not taking effect.. When I try to set it using "export JUNIT_TEST_CASE_PREFIX" locally on my test machine, it is working good

myadla:
Task Run Podified EDPM deployment failed running on host controller

myadla:
recheck

elfiesmelfie:
Thank you for making this investigation, unfortunately, we're going with the other option since we can't configure the name and classname like we want for polarian.

Other option: https://github.com/infrawatch/feature-verification-tests/pull/132 

Please close this PR

PR #137, opened 2024-08-05
update the revert patch command

mgirgisf:
its an enhancement to confirm the revert patch command.

The discussion : https://github.com/infrawatch/feature-verification-tests/pull/119#pullrequestreview-2171851815


elfiesmelfie (approved):
Makes sense!

PR #141, opened 2024-08-30
[telemetry_logging] Update the journal tests

elfiesmelfie:
* Update var names: identifiers_list -> journal_list
* Replace "| wc" in shell with check for stdout_lines | length in failure conditions
* Add additional failure conditions
    * if there are no journal entries, the rc is 0 and the number of lines 1 ("-- No entries --")
    * there should be nothing output to stderr
* fix syntax error on multi-line command (add |)
* use "become: true", to give the task the escalated priviledges it needs to access the journal
* rename the task to include the test ID, so that the custom_logger plugin logs the test result

### DIFF
--- a/roles/telemetry_logging/tasks/journal_tests.yml
+++ b/roles/telemetry_logging/tasks/journal_tests.yml
@@ -1,10 +1,15 @@
 ---
-- name: Get journals
+- name: |
+    Get journals {{ item }}
+    {{ journal_test_id }}
+  become: true
   ansible.builtin.shell:
-    cmd:
+    cmd: |
       tstamp=$(date -d '30 minute ago' "+%Y-%m-%d %H:%M:%S")
-      journalctl -t "{{ item }}" --no-pager -S "${tstamp}" | wc -l
+      journalctl -t "{{ item }}" --no-pager -S "${tstamp}"
   register: journal_wc
   changed_when: false
   failed_when:  
-    - journal_wc.stdout | int <= 1
+    - journal_wc.stdout_lines | length <= 1
+    - '"-- No entries --" in journal_wc.stdout'
+    - journal_wc.stderr_lines | length > 0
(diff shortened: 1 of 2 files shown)


elfiesmelfie:
This is tested in #140. I decided to add this PR so I could focus on getting PR#140 to gather the right log files (the files are there but not in the right place).

roles/telemetry_logging/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: "Verify journal entries - {{ journal_test_id }}"
+- name: "Verify journal entries"
```
ayefimov-1, line 2: How is the Polarian ID going to me assigned for this test?

ayefimov-1 (approved):
Looks good. Thx

myadla (commented):
looks good to me

myadla (approved):
lgtm

roles/telemetry_logging/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
...
   when: journal_list is defined
```
mgirgisf, line 3: where are we defining this "journal_list" exactly?

roles/telemetry_logging/tasks/journal_tests.yml
```diff
@@ -1,10 +1,15 @@
...
-    cmd:
+    cmd: |
       tstamp=$(date -d '30 minute ago' "+%Y-%m-%d %H:%M:%S")
-      journalctl -t "{{ item }}" --no-pager -S "${tstamp}" | wc -l
+      journalctl -t "{{ item }}" --no-pager -S "${tstamp}"
   register: journal_wc
   changed_when: false
   failed_when:  
-    - journal_wc.stdout | int <= 1
+    - journal_wc.stdout_lines | length <= 1
```
mgirgisf, line 13: Suggestion: These conditions are fine, but you might want to add debug output in case of failure to help with troubleshooting.

mgirgisf (approved):

PR #142, opened 2024-08-30
[ci] Run report_result in a separate hook

elfiesmelfie:
Report result should gather the results files and make them available on the controller.
If the callback plugin that is being used aggregates results at the end of the play, the this file is not available in the test playbook.

Use copy with remote src instead of shell/cp, so that the files can be copied from compute nodes into the controller logs dir Add a new play to make sure the output dir is created on the controller before copying the files.


ci/report_result.yml
```diff
@@ -1,18 +1,52 @@
...
+  gather_facts: true
+  vars_files:
+    - vars/common.yml
+  tasks:
+    - name: Collect the custom_logger results
+      delegate_to: controller
+      ansible.builtin.copy:
+        remote_src: true
+        src: "{{ ansible_env.HOME }}/test_run_result.out"
+        dest: "{{ logs_dir }}/{{ inventory_hostname }}-test_run_result.out"
```
mgirgisf, line 27: the log path should be : "{{ logs_dir }}/feature-verification-tests/"

I would suggest to add a var contains [ fvt_logs: "{{ logs_dir }}/feature-verification-tests/ " ]
that can be used across the tasks.

ci/report_result.yml
```diff
@@ -1,18 +1,52 @@
...
+    - compute
+  gather_facts: true
+  vars_files:
+    - vars/common.yml
+  tasks:
+    - name: Collect the custom_logger results
+      delegate_to: controller
+      ansible.builtin.copy:
+        remote_src: true
+        src: "{{ ansible_env.HOME }}/test_run_result.out"
```
mgirgisf, line 26: currently the callback_plugin is saving the files in "~/" not in the {{ ansible_env.HOME }} 

https://github.com/infrawatch/feature-verification-tests/blob/674716b9d70b4e16e5c5a524215e24b68576cc35/callback_plugins/custom_logger.py#L34

mgirgisf (changes requested):

ci/report_result.yml
elfiesmelfie: It's going to be the same in Zuul. But will not always be the same. The callback plugin will use the user's home directory on the ansible controller i.e. where ever the playbooks is called from (hooks in cifmw are run from the controller). It is possible that the you can run the playbook against localhost but with a different user. OR that this would be referenced from a target host.
This SHOULD be the same in our case, however, we could pass in the log dir via ansible.cfg. (might be a good extension to this later). As long as we don't reference the log files from the playbooks running on other hosts (like when grepping the results previously), we'll be fine.

mgirgisf (approved):
can you please remove this task https://github.com/infrawatch/feature-verification-tests/blob/77fd7c5731825ad55261eef6482e1bdd0f6482c4/ci/run_autoscaling_osp18.yml#L13 .

its creating unnecessary folder .

ci/report_result.yml
elfiesmelfie: Actually, we don't need to copy the files over from other hosts.
We just need to copy them to the logs_dir on the controller, from their other locations.

I think that there might be an issue when we're running multiple playbooks, since the output file will always have the same name, which will get overwritten by any subsequent tests. 
This is not going to be an issue immediately though, since all our jobs only run one test for the moment.

elfiesmelfie: For the case where we run multiple tests back-to-back, the custom logger should name the output files according to the play name OR append to an existing file instead of removing it (this could be a config option).

mgirgisf:
recheck

PR #143, opened 2024-09-02
Changing autoscaling tests names from RHOSHO-XXXX pattern to match tests polarion automation IDs

lnatapov:
Changing autoscaling tests names from RHOSHO-XXXX pattern to match tests polarion automation IDs


roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -87,7 +87,7 @@
   ansible.builtin.pause:
     minutes: 5
 
-- name: RHOSO-12664 Verify that the `cpu_alarm_low` alarm has been triggered
+- name: [TEST] verify_alaram_has_been_triggered
```
mgirgisf, line 90: the same I would say its better to add the alarm name

roles/telemetry_autoscaling/tasks/configure_heat.yml
```diff
@@ -27,19 +27,19 @@
...
   # TODO: Pass the parameters correctly. When I tried to pass the parameters
   # into the template.yaml file, they weren't passed to the instance initially.
-- name: RHOSO-12657 Configure heat template for automatically scaling instances
+- name: [TEST] configure heat template automatically scaling instances
   ansible.builtin.template:
     src: instance.yaml.j2
     dest: ~/templates/autoscaling/vnf/instance.yaml
 
-- name: RHOSO-12658 Create the resource to reference in the heat template
+- name: [TEST] create resource reference heat template
```
mgirgisf, line 35: I think removing "the, to , in the" making it a little bit difficult to understand what's meant by this task.
do we have a constrain on the words count ?

roles/telemetry_autoscaling/tasks/creating_stack.yml
```diff
@@ -69,15 +69,15 @@
...
+- name: [TEST] Verify that an instance was launched
   ansible.builtin.shell: |
     # source ~/overcloudrc;
     export STACK_ID=$({{ openstack_cmd }} stack show {{ stack_name }} -c id -f value);
     {{ openstack_cmd }} server list --long | grep $STACK_ID;
   register: result
   failed_when: result.rc >= 1
 
-- name: RHOSO-12652 Verify that the alarms were created for the stack
+- name: [TEST] Verify that alarm was created
```
mgirgisf, line 80: alarms were created , missing s.

roles/telemetry_autoscaling/tasks/main.yml
```diff
@@ -2,7 +2,7 @@
 # tasks file for telemetry_autoscaling
   # TODO: Update the pre-checks to include the resources that the stack expects to exist
   # i.e. networks, flavor, image, security group
-- name: RHOSO-12666 Test services are enabled for autoscaling
+- name: [TEST] Verify services running on overcloud
```
mgirgisf, line 5: we don't have overcloud anymore, so that can be confusing.

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -57,7 +57,7 @@
   register: busy_process
   with_items: "{{ vnf_instance_ip.stdout_lines }}"
 
-- name: RHOSO-12660 Verify that the `cpu_alarm_high` alarm has been triggered
+- name: [TEST] Verify alarm has been triggered
```
mgirgisf, line 60: I would say its better to add the alarm name, to verify which alarm we are talking about.

mgirgisf (changes requested):

roles/telemetry_autoscaling/tasks/configure_heat.yml
elfiesmelfie: This is to match the polarian test-automation-ids.
if the task is unclear, it is totally possible to make the task names multiple lines long to add some more information

mgirgisf:
this is failing with syntex error https://softwarefactory-project.io/zuul/t/rdoproject.org/build/2ebfadfe2f12480181256ad33ff3def6/log/controller/ci-framework-data/logs/post_deploy_run_functional_test.log

the name need to be between " " to accept the [ ] as string.

- name: "[TEST] Verify services running"

elfiesmelfie (commented):
I will update this to incorporate what i added to https://github.com/infrawatch/feature-verification-tests/pull/159/files

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -100,7 +100,9 @@
   register: result
   until: result.stdout == "alarm"
 
-- name: RHOSO-12663 Verify that the Orchestration service has scaled down the instances
+- name: |
+    TEST Verify Orchestration scaled down instances
+    Verify that the orchestration service has scaled down the instances
```
mgirgisf, line 105:
```suggestion
```

mgirgisf:
the problem is that we removed the logger prefix , so we should update the prefix there or check for the file exist before try to copy or check the result.

- name: "Check if the custom_logger results file exists"
  delegate_to: controller
  ansible.builtin.stat:
    path: "{{ ansible_env.HOME }}/test_run_result.out"
  register: file_stat

- name: "Collect the custom_logger results"
  delegate_to: controller
  ansible.builtin.copy:
    remote_src: true
    src: "{{ ansible_env.HOME }}/test_run_result.out"
    dest: "{{ logs_dir }}/test_run_result.out"
  when: file_stat.stat.exists

roles/telemetry_autoscaling/tasks/configure_heat.yml
```diff
@@ -27,19 +27,19 @@
 
   # TODO: Pass the parameters correctly. When I tried to pass the parameters
   # into the template.yaml file, they weren't passed to the instance initially.
-- name: RHOSO-12657 Configure heat template for automatically scaling instances
```
elfiesmelfie, line 30: For now, I think we need to keep the test IDs, since the custom logger is used to determine pass/fail, and it doesn't create a test_run_result.out file unless there are tests with test id prefix (i.e. RHOSO, RHOSP, UI)

elfiesmelfie: I'm going ot update this to match what is done in the other tests i.e. add a second line with the test id to the task name

mgirgisf (approved):

elfiesmelfie (approved):

PR #144, opened 2024-09-03
Metric verification job

vyzigold:
This adds tests to verify metrics from all our metric sources are in prometheus.

rh-jira: https://issues.redhat.com/browse/OSPRH-9694


ci/run_verify_metrics_osp18.yml
elfiesmelfie, line 1: The way we're adding jobs is changing a bit. Please have a look at https://github.com/infrawatch/feature-verification-tests/pull/140 to see some of the update that are needed.

roles/telemetry_verify_metrics/README.md
elfiesmelfie, line 1: Please update the docs for this role, and include the tests, the vars needed, etc.

roles/telemetry_verify_metrics/tasks/verify_ceilometer_metrics.yml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: Use openstack observabilityclient to verify ceilometer compute metrics are stored in prometheus
+  ansible.builtin.shell: |
+    {{ openstack_cmd }} metric show ceilometer_cpu
+  register: result
+  timeout: 30
+  retries: 5
+  until: result.rc == 0 and "ceilometer_cpu" in result.stdout
+
+- name: Delete the server
```
elfiesmelfie, line 35: These cleanup tasks should always be done.

Add a block/always here so that the cleanup will happen. 
You will also need to add some checks to make sure the resources exist before trying to delete them.

elfiesmelfie (commented):
I'll take a closer look at this tomorrow, when I have the other PRs lined up.

Please expect additional changes to the feedback and change requests as the in-flight patches continue to be merged and updated.

ci/run_verify_metrics_osp18.yml
vyzigold: Yes I'm aware of that. Main goal of the PR right now was to see this tested by Zuul and maybe get some early feedback. I'll mark this as a draft (which I wanted to do before, but forgot about it) until the required in-flight patches merge and I incorporate the needed changes.

vyzigold:
The job succeeded except for reporting the result, which I guess is gonna change a little? I'll get back to it in roughly 1.5 weeks.

roles/telemetry_verify_metrics/tasks/openstack_deployed.yml
elfiesmelfie, line 1: This is something we can put in a common check somewhere, since the autoscaling jobs does the same kind of checks[1].

There's an opportunity to combine this job with the autoscaling-fvt job, given the similarities (the autoscaling job is mostly separate at the moment because we patch the openstack versions).
OR
At the very least, add a task to the common role, which can be used to do this.


[1] https://github.com/infrawatch/feature-verification-tests/blob/master/roles/telemetry_autoscaling/tasks/test_services.yml

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+# tasks file for telemetry_verify_metrics
```
elfiesmelfie, line 2: If you follow the suggestion below to add the contents of this file to the common role, you can replace this task with:

```
- name: Verify OpenStack is deployed correctly
  ansible.builtin.include_role:
     name: common
     tasks_from: check_openstack_services_rhoso.yml
  vars:
        - [<CR>, <resource name>]
        - [<CR>, <resource_name>]
```

The data structure may be a list of dicts as well so it's easier to grok, but I think the list-of-lists might be easier to input.

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: Verify OpenStack is deployed correctly
+  ansible.builtin.include_tasks:
+    file: openstack_deployed.yml
+  tags: precheck
+
+- name: Verify RabbitMQ metrics are being exposed and stored
+  ansible.builtin.include_tasks:
+    file: verify_rabbitmq_metrics.yml
+  tags: test
+  when: '"rabbitmq" in metric_sources_to_test'
```
elfiesmelfie, line 12: Role vars should start with the role name i.e. telemetry_verify_metrics_metric_sources_to_test

roles/telemetry_verify_metrics/README.md
```diff
@@ -0,0 +1,50 @@
...
+    - Use openstack observabilityclient to verify RabbitMQ metrics are stored in Prometheus
+- Verify Ceilometer metrics are being exposed and stored
+    - Use openstack observabilityclient to verify Ceilometer central metrics are stored in Prometheus
+    - Use openstack observabilityclient to verify Ceilometer compute metrics are stored in Prometheus
+- Verify NodeExporter metrics are being exposed and stored
+    - Use openstack observabilityclient to verify NodeExporter metrics are stored in Prometheus
+
+Role Variables
+--------------
+openstack\_cmd - command to access openstack cli. For example: "oc rsh openstackclient openstack"
```
elfiesmelfie, line 32: metric_sources_to_test should be included here too, with the rename of the variable

elfiesmelfie (changes requested):

vyzigold:
I'd like to use the contents of https://github.com/infrawatch/feature-verification-tests/pull/157 to check for the ceilometer-compute-agent and node exporter containers on the computes. Afterwards this should be ready for another round of reviews.

vyzigold:
I think this is now a genuine failure of getting the Node Exporter metrics and the job is rightly detecting the failure. The job confirms, that ScrapeConfig exists and that the NodeExporter container on the compute node is healthy, but there just aren't any metrics in the Prometheus. This is further confirmed by the list of metrics gathered by must-gather: https://logserver.rdoproject.org/44/144/7aec4340af741ecd5d8bccda2c8e74036a4d3dee/github-check/functional-metric-verification-tests-osp18/c0b985d/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/ctlplane/ceilometer/metric_list

roles/common/tasks/container_test.yml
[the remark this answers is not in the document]
elfiesmelfie: You can use `delegate_to`. I would suggest having a second role though called telemetry_verify_metrics_compute. OR add the check to the verify metrics job using a second play in the playbook.

vyzigold:
recheck - the autoscaling fvt job failed on a tempest test??? `tempest.scenario.test_network_basic_ops.TestNetworkBasicOps` let's see if it passes on second try.

lnatapov:
Recheck

vyzigold:
recheck - issue should be fixed now

roles/common/tasks/container_test.yml
```diff
@@ -1,15 +1,27 @@
...
+
 - name: Get container status
   ansible.builtin.shell:
     cmd: |
-      podman ps -a --format "{{ '{{.Names}} {{.Status}}' }}" | grep {{ item }}
+      podman ps -a --format "{{ '{{.Names}} {{.Status}}' }}" | grep {{ container_name }}
   changed_when: false
   register: container_status
-
+  delegate_to: "{{ compute_node }}"
```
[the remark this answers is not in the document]
elfiesmelfie: The controller nodes have openshift. The compute nodes do not, so the services are running in containers there. 

One concern I have for this is that the container_status is set to the status on the last compute node IINM.
The state might be different on each of the nodes. I am not sure how to address this. 
@vyzigold Have you reviewed the logs for this task and seen what happens when the container is up on one compute but down on another?

roles/telemetry_verify_metrics/tasks/verify_rabbitmq_metrics.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: Check rabbitmq metric endpoint
+  ansible.builtin.shell: |
+    oc rsh openstackclient curl https://rabbitmq.openstack.svc:15691/metrics
+  register: result
+  changed_when: false
+  failed_when: result.rc >= 1
+
+- name: Check rabbitmq-cell1 metric endpoint
+  ansible.builtin.shell: |
+    oc rsh openstackclient curl https://rabbitmq-cell1.openstack.svc:15691/metrics
```
elfiesmelfie, line 30: Should this URL be available from outside the openstackclient pod?
Could you get the URL/IP address and make it more generic?
Is the route exposed? Or is the service URL available? (It's fine if it isn't, but if it is, we should try accessing it without the openstackclient pod)

elfiesmelfie (changes requested):

roles/common/tasks/container_test.yml
```diff
@@ -1,15 +1,27 @@
 ---
+- name: Set container name to a variable
```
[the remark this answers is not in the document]
vyzigold: Yes, putting the delegate_to into the telemetry_verify_metrics role is also a possibility. I'll do that.

roles/common/tasks/container_test.yml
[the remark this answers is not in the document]
vyzigold: This task loops through all compute nodes, so the results in the container_status variable is a list for each compute node. The task below then takes that list and checks results for each container.

roles/telemetry_verify_metrics/tasks/verify_rabbitmq_metrics.yml
vyzigold: We don't have any route, only a service. Regarding making it more generic, I guess I could get all the rabbitmq CRs and from the name of each I could construct the URL.

roles/common/tasks/check_openstack_services_rhoso.yml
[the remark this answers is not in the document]
vyzigold: The test_cr.yml file looks purely for the existence of a CR. I used it to check that the scrapeconfigs exist. This one also checks that the CRs have the correct status. (You can't checks status of scrapeconfigs as there is no status field)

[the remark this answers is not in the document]
vyzigold: So I think we need to test 2 different things:

1. Does the CR exist?
2. Does the CR have the correct state?

test_cr answers the first question, while the check_openstack_services_rhoso answers the second question. I think I can put this into a single file, it might become more understandable then. I'll have it soon.

roles/common/tasks/cr_tests.yml
```diff
@@ -0,0 +1,20 @@
...
+  ansible.builtin.command:
+    cmd: |
+      oc get {{ item.kind }} {{ item.name }} -o jsonpath='{.status.conditions[?(@.type=="{{ item.condition_type }}")].status}{"\n"}'
+  register: result
+  changed_when: false
+  failed_when:
+    - result.stdout != "True"
+  when:
+    - common_cr_ready_test_id is defined
+    - item.condition_type is defined
```
elfiesmelfie, line 20: Nicely done

roles/common/tasks/main.yml
```diff
@@ -35,10 +35,18 @@
...
+
 - name: "Verify container tests"
   when:
     - common_container_list is defined
     - common_container_test_id is defined
   ansible.builtin.include_tasks: "container_test.yml"
   loop: "{{ common_container_list }}"
-
+  loop_control:
+    loop_var: container_name
```
elfiesmelfie, line 52: Very nice, it is a stylistic change, but enhances readability

roles/telemetry_verify_metrics/tasks/check_compute_node_containers.yml
```diff
@@ -0,0 +1,7 @@
+- delegate_to: "{{ compute_node }}"
```
elfiesmelfie, line 1: I'm still not a fan of having this in the role, but I'm not going to block based on this.

I hope you'll later consider whether this is better being included in the ci/run_verify_metrics_osp18.yml playbook, as a separate play for compute and controller node tests.
Something similar (i.e. separating compute and controller tests into different plays) is done in the logging jobs.

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -0,0 +1,48 @@
...
+- name: Verify OpenStack is deployed correctly
+  ansible.builtin.include_role:
+    name: common
+  vars:
+    common_cr_test_id: RHOSO-1258
+    common_cr_ready_test_id: RHOSO-1259
+    common_cr_list:
+      - kind: telemetry
+        name: telemetry
+        condition_type: Ready
```
elfiesmelfie, line 11: I like the use of dicts here, which makes the test configs more readable than using lists.

roles/common/tasks/main.yml
vyzigold: Wasn't my idea. Ansible complained about "item" being redefined :rofl:

elfiesmelfie (approved):

mgirgisf (approved):

PR #147, opened 2024-09-10
[report_result] Update check for failed tests

elfiesmelfie:
grep fails when there are no lines matched (rc=1), but also when the file doesn't exist (rc>1). This change updates the check so that the error is reported properly.


mgirgisf (approved):

lnatapov (approved):
LGTM

PR #151, opened 2024-09-23
Add graphing cypress test

mgirgisf:


roles/telemetry_graphing/files/cypress.config.js
```diff
@@ -0,0 +1,8 @@
+const { defineConfig } = require('cypress')
```
csibbitt, line 1: I've not seen cypress before; this looks pretty neat

csibbitt (commented):
Looks like a good start! IIRC, the dashboards themselves render as SVGs, so in a future iteration, I think it could be reasonable to detect that certain graphs in the dashboards have non-0 signal lines in response to the load.

mgirgisf:
recheck

elfiesmelfie:
recheck

roles/telemetry_graphing/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: "[Test] Enable Graphing"
+  ansible.builtin.shell:
+    cmd: |
+      oc patch oscp/controlplane --type='json' -p '[{"op": "replace", "path": "/spec/telemetry/template/metricStorage/dashboardsEnabled", "value":true}]'
+  register: oscp_change
+  changed_when: false
+
```
elfiesmelfie, line 8: This should be in ci/run_graphing_test, as it is set-up for the test, rather than the test itself.

The change should also be reverted after (also not in the role)

A further note, you should wait for the change to be applied, in case it takes a while to resolve.
This should not be marked with "TEST" as it is setup.

roles/telemetry_graphing/vars/main.yml
elfiesmelfie, line 1: Are these supposed to be configurable? If not, this location is fine. If they should be configurable at any point, then put them in the defaults dir

roles/telemetry_graphing/tasks/cypress_test.yml
```diff
@@ -0,0 +1,87 @@
...
+    - name: Install dependencies
+      ansible.builtin.dnf:
+        name:
+          - xorg-x11-server-Xvfb
+          - gtk3-devel
+          - nss
+          - alsa-lib
+        state: present
+      become: true
+
```
elfiesmelfie, line 18: This is listed as a pre-req in the README. You should list it in the role description as something the role does, or remove it from here, and add it into the playbook in ci/

roles/telemetry_graphing/tasks/cypress_test.yml
```diff
@@ -0,0 +1,87 @@
...
+    chdir: "{{ cypress_tests_dir }}"
+  changed_when: false
+
+# RHOSO-12658 need to be updated
+- name: "[TEST] Run Cypress tests to validate dashboards exist RHOSO-12658"
+  ansible.builtin.shell: |
+      Xvfb :99 -ac &
+      export DISPLAY=:99
+      npx cypress run
+  args:
```
elfiesmelfie, line 69: If this functionality is being kept here, I suggest splitting this into multiple tasks, i.e. "install_cypress" and "run_cypress_test"

roles/telemetry_graphing/tasks/verify_url.yml
elfiesmelfie, line 1: I think that this would be better named "get_dashboard_url.yml" rather than verify url.
And should not set the URL unless it is accessible. (Or reset the URL to "" when the URL is not accessible)

When importing the task in main, use the `failed_when: openshift_console_url  is not defined`, so that the tests will only be run when the URL is accessible (i.e. defined)

elfiesmelfie (changes requested):

roles/telemetry_graphing/vars/main.yml
mgirgisf: No, its not, at least as far asI know for now.

roles/telemetry_graphing/tasks/cypress_test.yml
mgirgisf: I will update README file

roles/telemetry_graphing/tasks/cypress_test.yml
mgirgisf: this is not to install Cypress we already have another task to do so, but if I split it to multiple shell task it causing some problem in the scope of export. so I would say its better to kept that way

roles/telemetry_graphing/tasks/verify_url.yml
mgirgisf: No the dashboard_url is fixed somehow I just add this task to make sure if something change in the futur for how the console url is being triggered that we keep getting the right url but this task is to verify the access to the URL

callback_plugins/custom_logger.py
```diff
@@ -54,11 +54,13 @@ def playbook_on_stats(self, stats):
 
     def log_task_result(self, host, result, task_name):
         # test_run_result.out only interested in the test tasks, not setup or debug.
-        if "RHELOSP" in task_name or "RHOSO" in task_name:
+        if "RHELOSP" in task_name or "RHOSO" in task_name  or "UI" in task_name:
```
elfiesmelfie, line 57: This works, but is not scalable.
I want to add a new config options to allow adding the prefix via a config option, either as an env var or as a line in the ansible.cfg.

This is okay for now, or you can try adding a new update for it yourself.
There is a well-defined pattern, and you can use this PR as a reference: [https://github.com/infrawatch/feature-verification-tests/pull/160 (ignore the non-config stuff](https://github.com/infrawatch/feature-verification-tests/pull/160/commits/b98eedd69cb2285baa0b0f82b7f0fe997f374d39)

elfiesmelfie (approved):

ayefimov-1 (approved):
Looks okay to me.

PR #123, opened 2024-07-11
Merge qe_common role

ayefimov-1:
Merge QE Common role into master

PR #149, opened 2024-09-11
Add pod tests to logging job

elfiesmelfie:
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1115

PR #153, opened 2024-09-27
Add service tests to the logging job

myadla:
adding the tasks from https://github.com/infrawatch/feature-verification-tests/pull/123 into a spearate PR

https://issues.redhat.com/browse/OSPRH-9659


roles/common/tasks/container_status.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: "[TEST] verify container status"
```
elfiesmelfie, line 2: The TEST prefix isn't chosen yet, so this may change

roles/common/tasks/container_status.yml
```diff
@@ -0,0 +1,13 @@
...
+  ansible.builtin.shell:
+    cmd: |
+      podman ps -a --format "{{ '{{.Names}} {{.Status}}' }}" | grep "{{ container_name }}" | awk '{print $2}'
+  changed_when: false
```
elfiesmelfie, line 6: This needs a failed_when, if you're going to mark it with a TEST prefix, otherwise, it'll pass as long as the shell cmd returns something.

Alternatively, don't mark this task as a test, but mark the following task as a test.

Currently, we rely on the custom_logger to produce a file that gathers the test-ids=passed|failed output, so there should be a test ID in the second line of the task name.
Hopefully, we can remove this at some point soon, and add a check to report_result that checks the junit output instead.

roles/common/tasks/container_status.yml
```diff
@@ -0,0 +1,13 @@
...
+  register: container_status
+
+- name: Fail if container is not 'Up'
+  ansible.builtin.fail:
+    msg: "Container '{{ container_name }}' is not in 'Up' status. Current status: {{ container_status.stdout }}"
+  when:
+    - "'Up' not in container_status.stdout"
```
elfiesmelfie, line 13: I like this, it makes it clear why the test failed.

However, this task should be the one one with TEST in the name.

elfiesmelfie (changes requested):
As well as the comments inline, can you update the zuul job to run this? (based on https://github.com/infrawatch/feature-verification-tests/pull/149)

roles/common/README.md
```diff
@@ -0,0 +1,119 @@
+common
+=========
+
+The tests in this role are not specific to any one functional area but are an aggregate of common tests that can be used in all OSP 18.0/OCP jobs.
+
+Requirements
+------------
+
+None
```
elfiesmelfie, line 9: podman is required

roles/common/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Role Variables
+--------------
+Variable required for all tasks to run:
+
```
elfiesmelfie, line 14: You should only be adding the container tests

roles/common/tasks/container_status.yml
elfiesmelfie: The prefix will be "TEST"

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: "[TEST] verify status of multiple containers"
+  ansible.builtin.include_tasks:
+    file: container_status.yml
+  loop: "{{ containers }}"
+  loop_control:
+    label: "{{ item }}"
+  vars:
+    container_name: "{{ item }}"
+
+---
```
elfiesmelfie, line 12:
```suggestion
```

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+# All containers need to be up in order for this task to pass
+- name: "[TEST] verify status of multiple containers"
+  ansible.builtin.include_tasks:
+    file: container_status.yml
+  loop: "{{ containers }}"
+  loop_control:
+    label: "{{ item }}"
+  vars:
+    container_name: "{{ item }}"
+
```
elfiesmelfie, line 11: This is separate from adding the service checks. 

It should NOT be run unconditionally

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+---
+- name: "[TEST] Verify service for logging"
```
elfiesmelfie, line 13: There should be a conditional here, so that the service_tests.yml will only be included when the right vars are defined

roles/common/tasks/service_tests.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Verify service is running
+  ansible.builtin.shell:
+    cmd: |
+      oc get service -n "{{ service_nspace }}" "{{ service_name }}"
```
elfiesmelfie, line 5: To align with the rest of the tests, this should loop through a list of test items, so that multiple services can be tested together

elfiesmelfie (changes requested):

roles/common/tasks/container_status.yml
elfiesmelfie: We have discovered with other tests that this pattern won't work.

Since it's guarded by a "when", it won't run when the test case is passed.

elfiesmelfie:
This change was part of: https://github.com/infrawatch/feature-verification-tests/pull/144

PR #113, opened 2024-07-03
Merge Zuul Logging job

ayefimov-1:
Merge Logging job changes into Master

Depended-On: https://github.com/infrawatch/feature-verification-tests/pull/123
Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/124
#Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/118

PR #157, opened 2024-10-07
 Add container_test for Logging Job

mgirgisf:


ci/logging_tests_controller.yml
```diff
@@ -0,0 +1,184 @@
...
+      - repo-setup-edpm-deployment-logging
+      - run-os-edpm-deployment-logging
+      - ssh-known-hosts-edpm-deployment-logging
+      - telemetry-edpm-deployment-logging
+      - validate-network-edpm-deployment-logging
+  tasks:
+    - name: "Run pods completed tests"
+      ansible.builtin.import_role:
+        name: common
+
```
elfiesmelfie, line 184: You can add another play here to run the container tests OR add the container vars to an existing play (note the play name in line 46) to make the container tests run in the same play.

The reason that the pod vars are all split into different plays is that the pod lists are split by nspace and status string.

I will look into configuring the pod tests with a list like for the endpoint tests [1].

[1] https://github.com/infrawatch/feature-verification-tests/pull/113/files#diff-64b93d86b5903b16bb6cdee44d150891ead9ec2ac9383c8f001cb611fa7958e7R61-R77

roles/common/README.md
```diff
@@ -0,0 +1,73 @@
...
+      common_pod_test_id: "RHOSO-12752"
+      common_pod_status_str: "Running"
+      common_pod_nspace: openstack
+      common_pod_list:
+        - openstackclient
+    tasks:
+      - name: "Verify Running Pods"
+        ansible.builtin.import_role:
+          name: common
+
```
elfiesmelfie, line 63: Can you add in an example for the container tests and its vars?
Please also rename the vars to common_<var_name>, since they are part of the common role.

roles/telemetry_logging/tasks/main.yml
```diff
@@ -3,3 +3,10 @@
   when: journal_list is defined
   ansible.builtin.include_tasks: "journal_tests.yml"
   loop: "{{ journal_list }}"
+
+#All containers need to be up in order for this task to pass
+- name: "[TEST] verify status of multiple containers"
+  ansible.builtin.include_role:
+    name: common
+    tasks_from: container_status.yml
+  loop: "{{ container_list }}"
```
elfiesmelfie, line 12: If you add the include_tasks: container_tests to common/main.yml, you can skip the loop here, anmd ahve something like

```
- name: "TEST verify status of multiple containers"
  ansible.builtin.include_role:
      name: common
  vars:
      common_container_list:  "{{ container_list }}"
      
```

elfiesmelfie (changes requested):

roles/telemetry_logging/tasks/main.yml
```diff
@@ -3,3 +3,14 @@
...
+#All containers need to be up in order for this task to pass
+- name: "[TEST] verify status of multiple containers"
+  ansible.builtin.include_role:
+    name: common
+  vars:
+      common_container_test_id: "RHOSO-12753"
+      common_container_list:
+          - ceilometer_agent_compute
+          - ceilometer_agent_ipmi
+          - node_exporter
```
vyzigold, line 16: I don't see any connection of these containers to logging. ceilometer_agent_compute and node_exporter are deployed by the "telemetry" service. ceilometer_agent_ipmi will be deployed by the "telemetry power monitoring" service. You can deploy logging on the dataplane without these services. That means, that logging would work correctly, but these containers won't be there. If we want, we could check these containers in some general compute telemetry role, but not in a logging role. What we could check for logging on the computes is that rsyslog is installed and running, but I think rsyslog runs directly on the host, not inside a container.

vyzigold (changes requested):

roles/common/tasks/main.yml
ayefimov-1, line 1: Missing polarian testIDs in task name.

ayefimov-1 (changes requested):

ayefimov-1 (approved):
Looks good to me.

elfiesmelfie (approved):

vyzigold (approved):

PR #158, opened 2024-10-07
[logging] Add crd test to logging job

mgirgisf:


roles/common/tasks/crd_test.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Verify logging credentials "{{ item }}"
+  ansible.builtin.shell:
+    cmd: |
+      oc get crd "{{ item }}"
+  changed_when: false
+  register: output
+
+- name: |
+   [Test] Fail if CRD is not found
```
elfiesmelfie, line 10: I don't think this will work as expected. I haven't tested it, but it looks suspicious to me.
The status of this will either be "failed" or "skipped", never "pass"

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Verify pod - {{ common_pod_test_id }}"
+  when:
+    - common_pod_list is defined
+    - common_pod_nspace is defined
+    - common_pod_status_str is defined
+  ansible.builtin.include_tasks: "pod_tests.yml"
+  loop: "{{ common_pod_list }}"
```
elfiesmelfie, line 8: This should include a similar pattern for the CRD tests, where the tests are run conditionally, based on the vars existing

ci/logging_tests_controller.yml
```diff
@@ -0,0 +1,141 @@
...
+- name: "Verify the crds exist"
+  hosts: controller
+  gather_facts: no
+  ignore_errors: true
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  vars:
+    crd_test_id: "RHOSO-12670"
+    crd_list:
```
elfiesmelfie, line 131: These vars should be prefixed by the role name, since they belong to that role.
It's an ansible style thing that we haven't been enforcing so far, because it would need to be retroactively changed for old roles, but I would like to introduce it in new role that we're adding

ci/vars-logging-test.yml
```diff
@@ -1,8 +1,12 @@
 ---
-post_deploy_00_fvt_logging_computes:
-  source: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}/ci/logging_tests_computes.yml"
+# Skip running test_operator tests
+cifmw_test_operator_dry_run: true
```
elfiesmelfie, line 3: I don't think that we need this any longer. I _think_ that I added it to skip running the tempest tests when the tests were failing at some point.

roles/common/README.md
```diff
@@ -0,0 +1,77 @@
+common
+======
+
+The tests in this role are not specific to any one functional area but are an
+aggregate of common tests that can be used in all OSP 18.0/OCP jobs.
+
+The available tasks tests are:
+
+* pod tests
```
elfiesmelfie, line 9: Please update this to include the added tests.

The whole review will get easier once the pod tests merge, which I am trying to do ASAP.

roles/common/README.md
```diff
@@ -0,0 +1,77 @@
...
+For pod_tests.yml tasks:
+
+    common_pod_test_id
+    common_pod_list
+      - list of pods to validate
+    common_pod_status_str
+      - status of pods to check
+    common_pod_nspace
+      - list of projects where pods exist
+
```
elfiesmelfie, line 36: Add in the vars for the CRD tests

elfiesmelfie (changes requested):

mgirgisf:
recheck

ci/vars-logging-test.yml
vyzigold: @mgirgisf based on Emma's comment. Shouldn't these 2 lines be deleted?

roles/common/tasks/crd_tests.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Verify logging CRDD "{{ item }}"
```
vyzigold, line 2:
```suggestion
- name: Verify logging CRD "{{ item }}"
```

roles/common/tasks/main.yml
```diff
@@ -6,3 +6,10 @@
     - common_pod_status_str is defined
   ansible.builtin.include_tasks: "pod_tests.yml"
   loop: "{{ common_pod_list }}"
+
+- name: "Run crd tests "
```
vyzigold, line 10:
```suggestion
- name: "Run crd tests"
```

vyzigold (changes requested):

roles/common/tasks/crd_tests.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Verify logging CRD "{{ item }}"
```
ayefimov-1, line 2: multipe instances of issue with common role tasks should not have a "feature" name like "logging" associated with the task as they are suppose to be reusable and generic.

ci/logging_tests_controller.yml
```diff
@@ -119,3 +119,21 @@
...
+    PATH: "{{ cifmw_path }}"
+  vars:
+    common_crd_test_id: "RHOSO-12670"
+    common_crd_list:
+      - alertingrules.loki.grafana.com
+      - lokistacks.loki.grafana.com
+      - recordingrules.loki.grafana.com
+      - rulerconfigs.loki.grafana.com
+  tasks:
+    - name: "[TEST] verify the crds exist"
```
ayefimov-1, line 137: Multiple instances of missing polarian ID in tasks name as mapping of test failures is easier.

ayefimov-1 (changes requested):

roles/common/tasks/crd_tests.yml
ayefimov-1, line 2: Please remove "feature name" from task name as these are common tests for multiple jobs and feature. Suggest:

- name: Verify CRD "{{ item"

vyzigold (approved):

ayefimov-1 (approved):
Looks good.

roles/common/tasks/crd_tests.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Verify CRD "{{ item }}"
+  ansible.builtin.shell:
+    cmd: |
+      oc get crd "{{ item }}"
+  changed_when: false
+  register: output
+
+- name: Verify CRD is found "{{ common_crd_test_id }}"
```
elfiesmelfie, line 9: The test ID should be made optional OR this file should not be included unless it is defined (see comment below)

```suggestion
- name: Verify CRD is found {{ common_crd_test_id | default('') }}
```

elfiesmelfie (changes requested):

roles/common/tasks/main.yml
```diff
@@ -7,6 +7,13 @@
   ansible.builtin.include_tasks: "pod_tests.yml"
   loop: "{{ common_pod_list }}"
 
+
+- name: "Run crd tests"
+  when:
+    - common_crd_list is defined
```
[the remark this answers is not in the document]
mgirgisf: I updated the other tests also.

roles/common/tasks/pod_tests.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - block:
       # expects that one line will be returned
-      # todo: define what hapens when there are multiple pods returned
+      # tod: define what hapens when there are multiple pods returned
```
elfiesmelfie, line 4:
```suggestion
      # todo: define what happens when there are multiple pods returned
```

roles/common/README.md
```diff
@@ -90,6 +98,14 @@ can be set at the play level.
           common_project_list:
             - openshift-openstack-infra
             - openshift
+      - name: "Verify crd"
```
elfiesmelfie, line 101:
```suggestion

      - name: "Verify crd"
```

elfiesmelfie (changes requested):

elfiesmelfie (approved):

PR #160, opened 2024-10-08
[custom_junit] Make the test prefix configurable via ini file and env var

elfiesmelfie:
Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/132

### DIFF
--- a/callback_plugins/custom_junit.py
+++ b/callback_plugins/custom_junit.py
@@ -55,16 +55,20 @@ class CallbackModule(JunitCallbackModule):
     def mutate_task_name(self, task_name):
         print("enter mutate_task_name(task_name=%s)" % task_name)
 
+        if not self._test_case_prefix in task_name:
+            print("task_name (%s) does not contain prefix (%s)" % (task_name, self._test_case_prefix))
+
         new_name = task_name
         new_name = new_name.split("\n")[0]  # only use the first line, so we can include IDs and additional description
-        print(new_name)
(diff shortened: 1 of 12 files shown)


callback_plugins/custom_junit.py
```diff
@@ -6,6 +6,32 @@
 import re
 from ansible.utils._junit_xml import TestCase, TestError, TestFailure, TestSuite, TestSuites
 
+DOCUMENTATION = '''
+    callback: custom_junit
+    type: notification
+    short_description: TODO
+    description:
+      TODO
```
elfiesmelfie, line 14: custom_junit generates an XML file that polarion can read.
Only the tasks marked with $test_case_prefix are reported.
The first line of the task name (excluding the prefix) is converted to snake_case, and this becomes the testcase name in the results file.

callback_plugins/custom_junit.py
```diff
@@ -14,12 +40,15 @@ class CallbackModule(JunitCallbackModule):
     CALLBACK_NAME = 'custom_junit'
 
     def __init__(self):
+        self._defs = None
         super(CallbackModule, self).__init__()
+        self.set_options()
 
-        # Custom environment variable handling
         # Update this to parse these values from the config file, as well as the env.
         self._output_dir = os.path.expanduser("~/.ansible.log")
```
elfiesmelfie, line 48: This should also be configurable, so we could point it at the expect logs dir in zuul jobs.

PR #165, opened 2024-10-24
Switch to server_group for 18 autoscaling

vyzigold:
Latest updates to ceilometer and sg-core allow us to use server_group for autoscaling server grouping instead of the server_name_prefix. Unfortunately this functionality is only on ceilometer master, so we need to patch the openstackversions to use that. Because ceilometer-central-agent is running on the compute nodes, we need to redeploy the dataplane after patching the openstackversions.

I also found a few TODOs and other places that I thought I can enhance a little. I included that as well.

### DIFF
--- a/ci/run_autoscaling_osp18.yml
+++ b/ci/run_autoscaling_osp18.yml
@@ -29,6 +29,17 @@
           tags:
             - setup
 
+        - name: Redeploy the dataplane
+          ansible.builtin.shell:
+            cmd: |
+              oc get osdpd edpm-deployment -oyaml > /tmp/osdpd.yaml
+              oc delete osdpd edpm-deployment
+              sleep 10
+              oc apply -f /tmp/osdpd.yaml
+          when: "{{ patch_openstackversions | bool }}"
+          tags:
+            - setup
+
         - name: Patch observabilityclient into openstackclient
           ansible.builtin.shell:
             cmd: |
@@ -51,6 +62,18 @@
           tags:
             - setup
 
+        - name: Wait until the osdpd is redeployed to continue
+          ansible.builtin.shell:
+            cmd: |
+              oc get osdpd | grep "Setup complete"
+          retries: 60
+          delay: 45
+          until: output.stdout_lines | length == 1
+          register: output
+          when: "{{ patch_openstackversions | bool }}"
+          tags:
+            - setup
+
         - name: "Run Telemetry Autoscaling tests"
           ansible.builtin.import_role:
             name: telemetry_autoscaling
@@ -61,4 +84,15 @@
             cmd: |
               oc patch  openstackversions controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
           when: "{{ patch_openstackversions | bool }}"
+
+        - name: Redeploy the dataplane
+          ansible.builtin.shell:
+            cmd: |
+              oc get osdpd edpm-deployment -oyaml > /tmp/osdpd.yaml
+              oc delete osdpd edpm-deployment
+              sleep 10
+              oc apply -f /tmp/osdpd.yaml
+          when: "{{ patch_openstackversions | bool }}"
+          tags:
+            - setup
 
--- a/roles/telemetry_autoscaling/tasks/creating_stack.yml
+++ b/roles/telemetry_autoscaling/tasks/creating_stack.yml
@@ -63,7 +63,7 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} metric list
   register: result
-  failed_when: result.rc >= 1
+  failed_when: '"ceilometer_cpu" not in result.stdout'
 
 - name: Print the result
   ansible.builtin.debug:
@@ -82,7 +82,7 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} alarm list
   register: result
-  failed_when: result.rc >= 1
+  failed_when: '"cpu_alarm_high" not in result.stdout or "cpu_alarm_low" not in result.stdout'
 
 - name: Note the physical_resource_id values for the cpu_alarm_low resource
   ansible.builtin.shell: |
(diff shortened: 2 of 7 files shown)


roles/telemetry_autoscaling/tasks/creating_stack.yml
```diff
@@ -63,7 +63,7 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} metric list
   register: result
-  failed_when: result.rc >= 1
+  failed_when: '"ceilometer_cpu" not in result.stdout'
```
vyzigold, line 66: The original didn't do much checking. I think this condition is better as it does what's written in the task name.

roles/telemetry_autoscaling/tasks/creating_stack.yml
```diff
@@ -82,7 +82,7 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} alarm list
   register: result
-  failed_when: result.rc >= 1
+  failed_when: '"cpu_alarm_high" not in result.stdout or "cpu_alarm_low" not in result.stdout'
```
vyzigold, line 85: The original didn't do much checking. I think this condition is better as it does what's written in the task name.

roles/telemetry_autoscaling/tasks/creating_stack.yml
```diff
@@ -101,20 +101,19 @@
 - name: RHOSO-12653 Verify physical_resource_id match the alarm id for cpu_alarm_low
   ansible.builtin.shell: |
     # source ~/overcloudrc;
-    {{ openstack_cmd }} alarm list |grep -i cpu_alarm_low | awk '{print $2}'
+    {{ openstack_cmd }} alarm list |grep -i cpu_alarm_low | grep {{ stack_name }} | awk '{print $2}'
```
vyzigold, line 104: I added a grep for the stack_name. This is just to allow to run these tests when there is already an autoscaling stack with a different name present in the cluster.

roles/telemetry_autoscaling/tasks/creating_stack.yml
```diff
@@ -101,20 +101,19 @@
...
   register: alarm_id_low
   failed_when:
     - physical_resource_id_low.stdout != alarm_id_low.stdout
 
 - name: RHOSO-12654 Verify physical_resource_id match the alarm id for cpu_alarm_high
   ansible.builtin.shell: |
     # source ~/overcloudrc;
-    {{ openstack_cmd }} alarm list |grep -i cpu_alarm_high | awk '{print $2}'
+    {{ openstack_cmd }} alarm list |grep -i cpu_alarm_high | grep {{ stack_name }} | awk '{print $2}'
```
vyzigold, line 112: I added a grep for the stack_name. This is just to allow to run these tests when there is already an autoscaling stack with a different name present in the cluster.

roles/telemetry_autoscaling/tasks/post_teardown.yml
```diff
@@ -14,10 +14,9 @@
 
 - name: Wait until the stack is deleted
```
vyzigold, line 15: I don't think this ever worked correctly. In a situation when this should pass, the original output.stdout_lines would be an empty list, not 0. I checked a recent CI run and this actually failed. I added | wc -l so the check is now similar to a few other ones from this role.

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -62,6 +62,7 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} alarm list -c state -c name -f value| \
         grep -i "cpu_alarm_high" | \
+        grep -i "{{ stack_name }}" | \
```
vyzigold, line 65: I added a grep for the stack_name. This is just to allow to run these tests when there is already an autoscaling stack with a different name present in the cluster.

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -92,20 +93,18 @@
     # source ~/overcloudrc;
     {{ openstack_cmd }} alarm list -c state -c name -f value| \
         grep -i "cpu_alarm_low" | \
+        grep -i "{{ stack_name }}" | \
```
vyzigold, line 96: I added a grep for the stack_name. This is just to allow to run these tests when there is already an autoscaling stack with a different name present in the cluster.

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -92,20 +93,18 @@
...
         awk '{print $2}'
   retries: 100
   delay: 5
   register: result
   until: result.stdout == "alarm"
 
-  # TODO: the metering.server group metadata was used for gnocchi alarm
-  # selection.
-  # prom uses the instance name, so the metadata MIGHT be removed, and a new
-  # check for whether the scaling group has scaled down may be needed.
```
vyzigold, line 104: This is now actually the prefered way.

ci/run_autoscaling_osp18.yml
```diff
@@ -62,3 +85,14 @@
               oc patch  openstackversions controlplane  --type json -p='[{"op": "replace", "path": "/spec/customContainerImages", "value": {} }]'
           when: "{{ patch_openstackversions | bool }}"
 
+        - name: Redeploy the dataplane
```
elfiesmelfie, line 88: Why is the dataplane being redeployed twice?

roles/telemetry_autoscaling/tasks/post_teardown.yml
elfiesmelfie: This should never have worked. I suspect that the ``ignore_errors: true`` is also not needed, and may have been the reason that this wasn't causing issues

ci/run_autoscaling_osp18.yml
vyzigold: It's redeployed again to revert the changes done by the first redeployment. So after the first time we have master ceilometer containers. After the second time we have antelope containers again.

vyzigold: right

vyzigold: Dataplane is updated only as a reaction to a new OpenstackDataplaneDeployment CR. That's why there are tasks to copy the current osdpd into a file, delete the osdpd and apply it again from the file.

elfiesmelfie (approved):

vyzigold:
recheck graphing test failure.

PR #172, opened 2024-11-18
[ci] Set the cifmw_polarion_jump_result_dir to fvt report output

elfiesmelfie:


elfiesmelfie:
recheck

ci/vars-functional-test.yml
```diff
@@ -1,6 +1,7 @@
 ---
 # temp: skip os-net-setup
 cifmw_os_net_setup_config: []
+cifmw_polarion_jump_result_dir: “{{ ansible_user_dir }}/ci-framework-data/logs/feature-verification-tests”
```
elfiesmelfie, line 4: I'm not 100% convinced that this works from here. Also, we don't have polarion running here, so this should be closed IMHO

elfiesmelfie:
This looks like it is not needed so I will close this PR

PR #175, opened 2024-11-19
Determine API versions based on observability strategy

elfiesmelfie:
Depending on the observability strategy that STF uses, observability components will be installed from community or Red hat sources. These crate the same CRDs, but with different APIs e.g.
prometheusrules.monitoring.rhobs vs
prometheusrules.monitoring.coreos.com. An apiVersion needs to be set when creating these resources during the tests, and the tests will fail if the correct version is not available.

The test_alerts role will select which observability APIs to use based on which observability_straegy is set. The tests will also explicitly use the fully qualified name of the objects using ``oc get``.

The following roles are updated:
* test_alerts
* test_verify_email

Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/177

### DIFF
--- a/roles/test_alerts/tasks/get_observability_api.yml
+++ b/roles/test_alerts/tasks/get_observability_api.yml
@@ -1,0 +1,12 @@
+---
+- name: "Get the observability strategy"
+  ansible.builtin.shell:
+    cmd: |
+      oc get stf default -ojsonpath='{.spec.observabilityStrategy}'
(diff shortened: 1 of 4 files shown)


elfiesmelfie:
recheck

vkmc (approved):
LGTM, thanks!

mgirgisf (approved):

vkmc (approved):

PR #176, opened 2024-11-19
Rename tasks to have test id and description

elfiesmelfie:
The tasks were previously renamed to just be the test id, due to limitations in the custom_logger callbaack plugin. This was addressed, and the test_report_result reports in the correct format now, so the names can be reverted.
This makes reports more grokable, since the summary_result.log contains a list of the passed/failed tests with the task names.

Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/177

### DIFF
--- a/roles/test_collectd/tasks/main.yml
+++ b/roles/test_collectd/tasks/main.yml
@@ -1,6 +1,5 @@
 ---
-- name: RHELOSP-60406
-#     Description: Check whether collectd container is running
+- name: RHELOSP-60406 Check whether collectd container is running
   ansible.builtin.shell: |
     set -o pipefail
     {{ container_bin }} ps | grep {{ collectd_container_name }}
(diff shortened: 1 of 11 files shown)


roles/test_collectd/tasks/main.yml
ayefimov-1, line 1: Do the name strings have to be double qouted?  This file and qdr_test/tasks/main.yml file names are not double qouted!

elfiesmelfie: Unless they have some special characters, they don't have to be quoted.
I kept in quotes consistent with the file that the task is in

elfiesmelfie:
recheck

elfiesmelfie:
recheck

vkmc (approved):
So nice to be able to have the description with the ID... definitely easier to debug in the long run. LGTM, thanks!

PR #178, opened 2024-11-20
[ci] Update the test playbooks to ignore errors in tests

elfiesmelfie:
The tests should be run with ignore_errors: true in order to run all the tests. Without this, the tests will only run until the first failing test. This change is inline with the way the other test roles are included

### DIFF
--- a/ci/run_autoscaling_osp18.yml
+++ b/ci/run_autoscaling_osp18.yml
@@ -77,6 +77,7 @@
         - name: "Run Telemetry Autoscaling tests"
           ansible.builtin.import_role:
             name: telemetry_autoscaling
+          ignore_errors: true
 
       always:
         - name: Revert the version update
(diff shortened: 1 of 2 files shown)


lnatapov (approved):
lgtm

mgirgisf:
Its better if we add that to logging job also  https://github.com/infrawatch/feature-verification-tests/blob/a29eec287f61f95e07fac30c851070cc74da62ad/ci/logging_tests_computes.yml#L15

https://github.com/infrawatch/feature-verification-tests/blob/master/ci/logging_tests_controller.yml

elfiesmelfie:
> Its better if we add that to logging job also
> 
> https://github.com/infrawatch/feature-verification-tests/blob/a29eec287f61f95e07fac30c851070cc74da62ad/ci/logging_tests_computes.yml#L15
> 
> https://github.com/infrawatch/feature-verification-tests/blob/master/ci/logging_tests_controller.yml

It's set in the play.

vyzigold (approved):

PR #179, opened 2024-11-22
[test_alerts] Update the command for checking the alert in alert manager

elfiesmelfie:
The prometheus container in the prometheus-default pod no longer has wget included.
The equivalent curl command is used instead.


vkmc (approved):
Thanks Emma!

mgirgisf (approved):

PR #180, opened 2024-11-24
Fixing RHELOSP-176036 and RHELOSP-176036

lnatapov:
Fixing RHELOSP-176036 and RHELOSP-176036

### DIFF
--- a/roles/test_sensubility/tasks/test_health_status.yml
+++ b/roles/test_sensubility/tasks/test_health_status.yml
@@ -1,8 +1,8 @@
   #In this test we choose one container (logrotate_crond),checking that it is running 
   #and then stopping the container and starting it again and checking that its health status changes accordingly.
 
-- name: RHELOSP-176011 Check that logrotate_crond container is running on controller-0
-  delegate_to: "{{ groups['overcloud_nodes'][0] }}"
+- name: RHELOSP-176011 Check that logrotate_crond container is running on ceph-0
+  delegate_to: ceph-0
   become: true
   delegate_facts: True
   ansible.builtin.shell:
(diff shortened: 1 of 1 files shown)


roles/test_sensubility/tasks/test_health_status.yml
```diff
@@ -1,8 +1,8 @@
   #In this test we choose one container (logrotate_crond),checking that it is running 
   #and then stopping the container and starting it again and checking that its health status changes accordingly.
 
-- name: RHELOSP-176011 Check that logrotate_crond container is running on controller-0
-  delegate_to: "{{ groups['overcloud_nodes'][0] }}"
+- name: RHELOSP-176011 Check that logrotate_crond container is running on ceph-0
```
elfiesmelfie, line 4: Is this container expected to run on only ceph? or all nodes?

roles/test_sensubility/tasks/test_health_status.yml
```diff
@@ -1,8 +1,8 @@
...
+  delegate_to: ceph-0
```
elfiesmelfie, line 5: All of these are delegated to ceph-0. Does it matter what host it runs on, or can it run on any host?

It might make more sense to remove the delegate_to and update the stf playbook to add a play that runs on the ceph node for this. Could this also run on the controller?

lnatapov: yes,it can run on all overcloud nodes. They all have that container

roles/test_sensubility/tasks/test_health_status.yml
lnatapov: all

mgirgisf (commented):
could you please add a comment or commit message for that overcloud_nodes[0] and ceph-0 are running on the same host

elfiesmelfie (approved):

mgirgisf (approved):

vkmc:
Thanks for adding that clarification Marihan! Agree that having this in the commit message would make reviews easier. @lnatapov could you add it so we keep it in the history for future reference?

vkmc (approved):

elfiesmelfie:
recheck

PR #184, opened 2024-11-28
Replace wc -l with stdout_lines | length

elfiesmelfie:
Updated roles:
* test_alert
* test_snmp_trap
* test_metrics_retention
* test_sensubility
* test_verify_email

### DIFF
--- a/roles/test_alerts/tasks/test_creating_a_standard_alert_route_in_alert_manager.yml
+++ b/roles/test_alerts/tasks/test_creating_a_standard_alert_route_in_alert_manager.yml
@@ -45,18 +45,18 @@
     cmd: >-
       oc exec -it prometheus-default-0 -c prometheus -- /bin/sh -c 'wget --header \
         "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-        https://default-alertmanager-proxy:9095/api/v1/alerts -q -O -' | grep 'active' | grep 'Collectd metrics receive rate is zero' | wc -l
+        https://default-alertmanager-proxy:9095/api/v1/alerts -q -O -' | grep 'active' | grep 'Collectd metrics receive rate is zero'
   register: cmd_output
   changed_when: false
-  failed_when: cmd_output.stdout|int == 0
+  failed_when: cmd_output.stdout_lines | length == 0
 
 - name: "RHELOSP-148699 Verify that the alert is firing in Prometheus"
   ansible.builtin.shell:
     cmd: >-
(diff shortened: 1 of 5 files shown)


mgirgisf (dismissed):
thanks emma

roles/test_sensubility/tasks/test_health_status.yml
```diff
@@ -71,9 +66,8 @@
     cmd: >-
       /usr/bin/curl -k {{ prom_auth_string }} \
         -g https://{{ prom_url }}/api/v1/query? \
-        --data-urlencode 'query=last_over_time(sensubility_container_health_status{process="logrotate_crond",host="controller-0.redhat.local"}[10m])' \
-        | grep -oP '(?<="value":).*' | awk -F, '{ print $2 }' |  grep -o '[0-9]\+' | grep 1 | wc -l
+        --data-urlencode 'query=last_over_time(sensubility_container_health_status{process="logrotate_crond",host="ceph-0.redhat.local"}[10m])' \
```
vkmc, line 69: This change seems unrelated

vkmc (changes requested):
Not sure if the change in the host for the RHELOSP-176035 is intended, but it seems unrelated. Maybe this comes from a different change?

roles/test_sensubility/tasks/test_health_status.yml
elfiesmelfie: This comes from here: https://github.com/infrawatch/feature-verification-tests/pull/180

I believe this shows up because of the order in which these changes were proposed.

elfiesmelfie: I've targetted this change against PR#180, instead of master. I think that should reflect the correct changes for this PR, as well as represent the dependency

vkmc (approved):
LGTM, thanks Emma!

roles/test_alerts/tasks/test_creating_a_standard_alert_route_in_alert_manager.yml
```diff
@@ -45,18 +45,19 @@
     cmd: >-
       oc exec -it prometheus-default-0 -c prometheus -- /bin/sh -c 'curl -k -H \
         "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-        https://default-alertmanager-proxy:9095/api/v1/alerts' | grep 'active' | grep 'Collectd metrics receive rate is zero' | wc -l
+        https://default-alertmanager-proxy:9095/api/v1/alerts' | grep 'active' | grep 'Collectd metrics receive rate is zero'
+
```
elfiesmelfie, line 49:
```suggestion
```

ayefimov-1 (approved):
Looks like better logic.

lnatapov (approved):
LGTM

PR #187, opened 2024-12-05
Add block/always to tests to separate test procedure from cleanup

elfiesmelfie:
This let the cleanup always be done, even when there are test failures.


lnatapov (approved):
LGTM

mgirgisf (approved):

PR #192, opened 2024-12-06
[graphing] Polarian Report

mgirgisf:
Replace fail with assert module to have state of Pass or Fail instead of Skip.

### DIFF
--- a/roles/telemetry_graphing/tasks/cypress_test.yml
+++ b/roles/telemetry_graphing/tasks/cypress_test.yml
@@ -83,8 +83,10 @@
     var: cypress_run.stdout_lines
 
 - name: |
-    TEST Check if all tests passed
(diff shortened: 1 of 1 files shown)


roles/telemetry_graphing/tasks/cypress_test.yml
```diff
@@ -83,8 +83,10 @@
     var: cypress_run.stdout_lines
 
 - name: |
-    TEST Check if all tests passed
+    TEST Check if all Cypress tests passed
```
elfiesmelfie, line 86:
```suggestion
    TEST Check if all tests passed
```

elfiesmelfie (changes requested):

elfiesmelfie (approved):

lnatapov (approved):
LGTM

PR #202, opened 2024-12-20
Fix logging after update to CLO 6

vyzigold:
In FR1 we stopped using the dataplane-operator. It was merged into the openstack-operator, so we can't expect it to be running. Check that openstack-operator is running instead.

Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/572

### DIFF
--- a/ci/logging_tests_controller.yml
+++ b/ci/logging_tests_controller.yml
@@ -12,7 +12,7 @@
     common_pod_nspace: openstack-operators
     common_pod_list:
       - telemetry-operator-controller-manager
-      - dataplane-operator-controller-manager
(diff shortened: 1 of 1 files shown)


vyzigold:
There seem to be other issues, but that's next year problem I think.

vyzigold:
So the issue from last year is, that the logging-view-plugin isn't running.

vyzigold:
And Juan confirmed, that it's the expected behavior for the pod to not be there in CLO 6. I removed the check.

elfiesmelfie (approved):

vkmc (approved):
LGTM

PR #203, opened 2025-01-16
[common] Use a config dict for configuring tests

elfiesmelfie:
Allow a list of key/value pairs to be passed to configure the tests.
This allows for passing test configs that don't share all common attributes e.g. run the pod tests to check multiple namespaces in one go


elfiesmelfie:
recheck

PR #207, opened 2025-01-22
[common] Update the CR tests to mark one as optional

elfiesmelfie:
Only include the test_id if the condition_type is defined. Set a default for the condition_type so that the task name is rendered correctly in output plugins.

### DIFF
--- a/roles/common/tasks/cr_tests.yml
+++ b/roles/common/tasks/cr_tests.yml
@@ -18,7 +18,7 @@
     condition_type: "{{ item.condition_type if item.condition_type is defined else '' }}"
 
 - name: |
-    {{ "TEST " if condition_type | length >0 }}Verify that the {{ item.kind }} {{ item.name }} CR is {{ condition_type }}
+    {{ 'TEST ' if condition_type | length > 0 }}Verify that the {{ item.kind }} {{ item.name }} CR is {{ condition_type }}
     {{ common_cr_ready_test_id if condition_type | length > 0 }}
   ansible.builtin.command:
     cmd: |
@@ -27,3 +27,6 @@
   changed_when: false
   failed_when:
     - result.stdout != "True"
+  when:
+    - common_cr_ready_test_id is defined
+    - condition_type | length > 0
--- a/roles/common/tasks/cr_tests.yml
+++ b/roles/common/tasks/cr_tests.yml
@@ -9,13 +9,21 @@
   failed_when:
     - result.rc != 0
 
-- name: Verify that a CR is ready
+  # This is added so that the task name will be rendered correctly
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
recheck

roles/common/tasks/cr_tests.yml
```diff
@@ -7,14 +10,21 @@
...
+  # This is added so that the task name will be rendered correctly
+  # If item.condition_type is used and it doesn't exist, then the
+  # literal template text will appear as the task name
+  # e.g. "Verify that the {{ item.kind }} {{ item.name }} CR is {{ item.condition_type }}"
+- name: Set the condition type
+  ansible.builtin.set_fact:
+    condition_type: "{{ item.condition_type if item.condition_type is defined else '' }}"
+
+- name: |
+    Verify that the {{ item.kind }} {{ item.name }} CR is {{ condition_type }}
```
vyzigold, line 22: The name makes less sense to me like this. It's expected for the condition_type to not be defined (to be set to '' by the task above) right? In that case the name will render into just "Verify that the <kind> <name> CR is".

The condition type is used to determine where to look in a CR to determine it's in some expected state. A not so practical example could be, that we want to check that the CeilometerReady is True in the Telemetry CR. Now the name would be: "Verify that the telemetry telemetry CR is CeilometerReady".

You could reword it to something like: "Verify that the {{ condition_type }} condition of {{ item.kind }} {{ item.name }} is True"

roles/common/tasks/cr_tests.yml
```diff
@@ -7,14 +10,21 @@
...
     cmd: |
-      oc get {{ item.kind }} {{ item.name }} -o jsonpath='{.status.conditions[?(@.type=="{{ item.condition_type }}")].status}{"\n"}'
+      oc get {{ item.kind }} {{ item.name }} -o jsonpath='{.status.conditions[?(@.type=="{{ condition_type }}")].status}{"\n"}'
   register: result
   changed_when: false
   failed_when:
     - result.stdout != "True"
-  when:
-    - common_cr_ready_test_id is defined
-    - item.condition_type is defined
```
vyzigold, line 20: There are CRs, which don't have status conditions to check, so in that case the cr_test would just check their existence (the top task of this file). What happens with the when removed? I think it just fails in that case right?

vyzigold (changes requested):

roles/common/tasks/cr_tests.yml
elfiesmelfie: I don't have an answer for that. Looking at this here, I think the when should be re-added. I can't remember why it was removed.

vyzigold (dismissed):

PR #209, opened 2025-01-23
[test_snmp_traps] Add a task to get prom creds at the start of the test

elfiesmelfie:
This ensures that the prom creds are available for the test.
Previously, the tests suites running this would "assume" that the vars were already set.


ayefimov-1 (approved):
Looks good.

vkmc (approved):
LGTM, thanks!

PR #211, opened 2025-01-24
Make graphing test non-voting

ayefimov-1:
Make graphing job non-voting

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -112,6 +112,6 @@
             irrelevant-files: *irrelevant_files
         - functional-tests-on-osp18
         - functional-logging-tests-osp18
-        - functional-graphing-tests-osp18
+        - functional-graphing-tests-osp18:
             voting: false
         - functional-metric-verification-tests-osp18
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -88,6 +88,7 @@
 - job:
     name: functional-graphing-tests-osp18
     parent: telemetry-operator-multinode-autoscaling
+    voting: false
```
elfiesmelfie, line 91: Technically, this is correct. 

However, setting the value here will set it for any project that uses this job, so setting the value here is not considered best practice.

In general, one would set this per-project as an anonymous job variant instead
i.e.
Line 116 would become: 

```
- functional-graphing-tests-osp18:
    voting: false
```

elfiesmelfie (changes requested):

.zuul.yaml
ayefimov-1: Fixed

elfiesmelfie (approved):

mgirgisf (approved):

vyzigold (approved):

mrunge:
This is superseded by https://github.com/infrawatch/feature-verification-tests/pull/213 ?

elfiesmelfie:
> This is superseded by #213 ?

Yes, the change to make the job non-voting was a short-term workaround, which is not needed since #213 actually fixes it.

PR #218, opened 2025-01-29
Move container image patch to post_ctlplane hook

vyzigold:
If we manage to change the openstackversions as needed before the dataplane gets deployed the first time, we don't need to redeploy it later. This could speed up the job by like 30 minutes as well as fix our current issues.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -10,6 +10,7 @@
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir  }}/scenarios/centos-9/multinode-ci.yml"
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/telemetry-operator'].src_dir }}/ci/vars-autoscaling.yml"
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}/ci/vars-functional-test.yml"
+        - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/infrawatch/feature-verification-tests'].src_dir }}/ci/vars-use-master-containers.yml"
     roles:
       - zuul: github.com/openstack-k8s-operators/ci-framework
     required-projects: &required_projects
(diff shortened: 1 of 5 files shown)


vyzigold:
This kindof worked? From the must-gather I see we have master images in the controlplane and in the dataplane. But the queried metric doesn't seem to be in Prometheus possibly because the right labels are missing. This would suggest antelope image for ceilometer-compute-agent. I'll try to look further.

ci/vars-use-master-containers.yml
elfiesmelfie, line 1: Oh excellent, since this is a separate vars file, it will not effect downstream jobs that "inherit" from this one

roles/telemetry_autoscaling/tasks/test_autoscaling.yml
```diff
@@ -59,6 +59,19 @@
   register: busy_process
   with_items: "{{ vnf_instance_ip.stdout_lines }}"
 
+- name: Show queried metric
```
elfiesmelfie, line 62: As far as I know, this should work, since the update to install observabilityclient into the openstackclient is still there.

vyzigold: Yes. Ideally we have all the new observabilityclient patches when running the role. This is taken care of by the playbook in ci/ upstream and by backports downstream. However this doesn't actually show anything. I'll submit another version, which actually displays the output of the command. I think this can be quite valuable for future debugging.

vyzigold:
This worked the second time. I don't know what the issue was the first time, but it seems to succeed at least sometime. I'll refine the debug tasks a little, so that they can be used to give more inside if there is a similar failure in the future.

vyzigold:
Downside of this is, that we can't revert the containers back to antelope after testing. That would involve a dataplane redeploy. There is no way around it.

jlarriba (approved):

.zuul.yaml
```diff
@@ -112,7 +113,7 @@
             override-checkout: main
             irrelevant-files: *irrelevant_files
         - functional-tests-on-osp18:
-            voting: false
+            voting: true
```
mgirgisf, line 116: this is not required, as its true by default.

mgirgisf (approved):

PR #193, opened 2024-12-06
Add metrics verification tasks for power monitoring components

yadneshk:
Closes: [OSPRH-11638 ](https://issues.redhat.com/browse/OSPRH-11638)

PR #224, opened 2025-02-12
Fix kepler metrics tests

rsafrono:
Kepler metrics tests were added in [1] though the merged commit was not the correct one. This PR applies changes from [2] that had more correct code of the tests.

[1] https://github.com/infrawatch/feature-verification-tests/pull/193
[2] https://github.com/infrawatch/feature-verification-tests/commit/0b5d856fb33ce05cb5ff93dd67932a473318946d


roles/telemetry_verify_metrics/tasks/verify_kepler_metrics.yml
```diff
@@ -6,12 +6,6 @@
     common_cr_list:
       - kind: scrapeconfigs.monitoring.rhobs
         name: telemetry-kepler
```
elfiesmelfie, line 8: It looks like this wasn't used, so removing it is okay

roles/telemetry_verify_metrics/tasks/verify_kepler_metrics.yml
```diff
@@ -28,9 +22,15 @@
...
   ansible.builtin.shell: |
-    {{ openstack_cmd }} metric show --disable-rbac kepler_node_info
+    {{ openstack_cmd }} metric show --disable-rbac {{ item }}
   register: result
   delay: 30
   retries: 10
   changed_when: false
-  until: result.rc == 0 and "kepler_node_info" in result.stdout
+  until: result.rc == 0 and item in result.stdout
+  with_items:
```
elfiesmelfie, line 31: Is there a reason to specify this here rather then re-use the var from above? Perhaps with a different name

rsafrono: re-used the var from above. Just renamed it from kepler_dashboard_metrics to  kepler_metrics

roles/telemetry_verify_metrics/tasks/verify_kepler_metrics.yml
```diff
@@ -28,9 +22,15 @@
...
     RHOSO-1216
   ansible.builtin.shell: |
-    {{ openstack_cmd }} metric show --disable-rbac kepler_node_info
+    {{ openstack_cmd }} metric show --disable-rbac {{ item }}
   register: result
   delay: 30
   retries: 10
   changed_when: false
-  until: result.rc == 0 and "kepler_node_info" in result.stdout
+  until: result.rc == 0 and item in result.stdout
```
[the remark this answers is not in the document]
rsafrono: The suggested change did not work for me. It fails with this:
[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: result.rc == 0 and {{ item }} in result.stdout
fatal: [localhost]: FAILED! => {
    "msg": "The conditional check 'result.rc == 0 and {{ item }} in result.stdout' failed. The error was: error while evaluating conditional (result.rc == 0 and {{ item }} in result.stdout): Unable to look up a name or access an attribute in template string ({% if result.rc == 0 and kepler_vm_platform_joules_total in result.stdout %} True {% else %} False {% endif %}).\nMake sure your variable name does not contain invalid characters like '-': 'in <string>' requires string as left operand, not AnsibleUndefined. 'in <string>' requires string as left operand, not AnsibleUndefined. Unable to look up a name or access an attribute in template string ({% if result.rc == 0 and kepler_vm_platform_joules_total in result.stdout %} True {% else %} False {% endif %}).\nMake sure your variable name does not contain invalid characters like '-': 'in <string>' requires string as left operand, not AnsibleUndefined. 'in <string>' requires string as left operand, not AnsibleUndefined"
}

At the same time the original works as expected, tested also when only half of the condition satisfied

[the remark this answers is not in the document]
elfiesmelfie: Ah, now I understand why that is, the condition is dropped into a template already.

Thank you for checking that out and sharing the results of the testing

elfiesmelfie (approved):

vyzigold (approved):

PR #227, opened 2025-02-19
Adding a test that checks that collectd memory limit set to 512Mb

lnatapov:
Adding a test that checks that collectd memory limit set to 512Mb


roles/test_collectd/tasks/main.yml
```diff
@@ -28,3 +28,11 @@
   register: stat
   changed_when: false
   failed_when: stat.stdout_lines|length == 0
+
+- name: TEST Check the collectd memory limit is set to 512Mb
+  ansible.builtin.shell: |
+    {{ container_bin }} inspect {{ collectd_container_name }} | grep 512 -c
```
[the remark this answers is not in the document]
elfiesmelfie: The memory limit is represented in B.

A container run with a memory limit of 512M will actually fail this test.

Run a container and set the memory limit to 512M:

`podman run --memory 512m --name not_collectd hello:latest `

Then find the name 
Check the memory limit.
```
podman inspect not_collectd | grep Memory
               "Memory": 536870912,
               "KernelMemory": 0,
               "MemoryReservation": 0,
               "MemorySwap": 1073741824,
               "MemorySwappiness": -1,
```

```podman inspect not_collectd | grep  512```
will return a result for the command used to run the container, possibly also some hashes.
If the container was mis-managed, or deployed with the wrong settings, or even with some other property that includes "512", this test will pass, without verifying anything. Or, it might return multiple lines of result.

Podman allows updating the properties of a container, including the memory limits, so even detecting the options used to deploy a container may not represent the properties.

I suggest either modifying the grep command, or using jq (which we might already use elsewhere in the tests.

For a modified grep command, try 
```
{{ container_bin }} inspect {{ collectd_container_name }} | grep \"Memory\"
```

Which will match something like this: 
```
               "Memory": 536870912,
```

The fail condition for the ansible task can be:
 ```
failed_when: "536870912" not in memory.stdout
```

elfiesmelfie (changes requested):

roles/test_collectd/tasks/main.yml
```diff
@@ -28,3 +28,11 @@
   register: stat
   changed_when: false
   failed_when: stat.stdout_lines|length == 0
+
+- name: TEST Check collectd memory limit
```
elfiesmelfie, line 32: Looks good! This is fine as-is, however, if it was to fail, I think it might be useful to have a tasks before this to check for whether the container was created with a memory limit. 

```
podman inspect not_collectd4 | jq '.[].Config.CreateCommand'
[
  "podman",
  "run",
  "-itd",
  "--memory",
  "512m",
  "--name",
  "not_collectd4",
  "alpine"
]

$ podman inspect not_collectd4 | jq '.[].HostConfig.Memory'
536870912

```

I'm not sure if we expect jq to be installed for this set of tests. If we don't expect it to be installed, you can disregard this feedback.

If it is already installed, then you can follow-up in another PR, if you want to extend the test in this way.

elfiesmelfie (approved):
Looks good.
I have some suggestions for optional improvements that could be included in a follow-up.

IMHO, adding the task to check for whether a memory limit was configured would be useful for the cases where the current task fails, and would reduce the time needed to troubleshoot the issue.

roles/test_collectd/tasks/main.yml
```diff
@@ -28,3 +28,11 @@
...
+  ansible.builtin.shell: |
+    {{ container_bin }} inspect {{ collectd_container_name }} | grep -i \"Memory\"
+  register: memory
+  changed_when: false
+  failed_when: memory.stdout_lines != 536870912
```
paramite, line 37: I'm not 100% sure how exactly failed_when works, but if you grep on `Memory` then one line of the stdout will be `Memory: 536870912` , correct? I wonder then how this can work :). Does the `!=` works as `!contains`?

elfiesmelfie, line 37:
```suggestion
  failed_when: not "536870912" in memory.stdout
```

elfiesmelfie (changes requested):

elfiesmelfie (approved):

roles/test_collectd/tasks/main.yml
```diff
@@ -28,3 +28,11 @@
   register: stat
   changed_when: false
   failed_when: stat.stdout_lines|length == 0
+
+- name: TEST Check collectd memory limit
+  ansible.builtin.shell: |
+    {{ container_bin }} inspect {{ collectd_container_name }} | grep -i \"Memory\"
+  register: memory
+  changed_when: false
+  failed_when: not "536870912" in memory.stdout
```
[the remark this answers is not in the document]
mrunge: it is set to 512MB by default. You can change it in the deployment, but in that case the test would fail.

PR #230, opened 2025-02-25
Exclude OWNERS* from triggering fvt

jlarriba:


.zuul.yaml
```diff
@@ -34,6 +34,7 @@
       - ci/github/.*
       - callback_plugins/.*
       - README*
+      - OWNERS*
```
elfiesmelfie, line 37: I think this should be in telemetry-operator, since we don't have an OWNERS file here.

I will test this using a draft PR to telemetry-operator which depends on this PR.
This will confirm whether or not this will work

elfiesmelfie: Okay, so that worked, HOWEVER, no jobs at all were triggered.

This is an issue because the Zuul status is required for the PR to merge..
We have used a no-op job in the past, which serves only to return a passing status from Zuul.

I will add the feature-verification-tests-noop job to the telemetry-operator repo.
It will be triggered whenever an irrelevant file is changed.
It does nothing, and uses no resources that would effect the other jobs, so this is pretty safe.

PR #232, opened 2025-03-03
Fix rabbitmq test

vyzigold:


jlarriba (approved):

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -15,10 +15,10 @@
       - kind: ceilometer
         name: ceilometer
         condition_type: Ready
-      - kind: rabbitmq
+      - kind: rabbitmqs.rabbitmq.openstack.org
```
fmount, line 18: While this is good and it will fix main, I'm wondering about the implications of this change wrt previous releases.
I see we don't follow a FRx branching strategy, which I think it makes sense for this project, but because we don't have `rabbitmqs` in the previous releases (e.g. FR1), I'm not sure we should improve this code to check both CRDs and see which one has a Rabbit object.
One idea might be to simply extend this check to support both CRDs, unless we don't care about FR1 anymore as long as we branch FR2.
@jlarriba what do you think? This change is sufficient to fix main and test FR2 (that is going to have a branch today), but it might break FR1 because `kind: rabbitmqs.rabbitmq.openstack.org` isn't there yet. How do we deal with this kind of interface changes?

stuggi: both versions have `rabbitmqclusters`, which is the actual object handling rabbitmq deployment. could just check for them?

vyzigold: Hmm, you're right Francesco, we probably want to be able to test both for now. I'll rewrite this to use `rabbitmqclusters` as Martin suggested. This should get the tests into a working state for now and we can figure out what we want to do in the future later.

fmount: right, thank you both for the help with this. I didn't have a fresh environment to check my suggestions/assumptions and I think it's better to fallback to `rabbitmqclusters` if it resolves the group we're checking. It works for previous releases and we should be able to solve this problem for now.

jlarriba: Its my understanding that there will be no more patches into FR1 as soon as FR2 is released, right? So that is no concern.

vyzigold: To complete the conversation. Since we don't plan to test FR1 once FR2 releases, I created a jira to track the work to switch the kind from the current `rabbitmqclusters` to `rabbitmqs`. If we do that after FR2 releases, I think we should be fine and it seems as the easiest option.
https://issues.redhat.com/browse/OSPRH-14451

PR #234, opened 2025-03-06
Add test for health monitoring feature

paramite:


paramite:
Seems like FVT for autoscaling failed.

paramite:
recheck

elfiesmelfie (changes requested):
For now, it looks like we need to keep the RHOSO-XXXX in the tests.

I'll look into removing this dependency as soon as I can.
For reporting pass/fail, we look at the file generated by the  custom_logger, which reports tasks which have a test id with a particular-prefix.
For the XML reporting, we don't require this test id.

To get accurate reporting from the job, please add these RHOSO-XXX (XXX is some arbitrary number) so that the fail will be reported.

paramite:
Seems like we're good to go, right?

lnatapov (approved):
LGTM

paramite:
@elfiesmelfie Hey Emma, is there anything else that needs to be done in this PR?

PR #238, opened 2025-03-11
Add basic tests for VM power monitoring metrics

rsafrono:
Added 2 tests:
Test idle VM power monitoring metrics are not zero 
Test dynamic VM power monitoring metrics are increasing

### DIFF
--- a/roles/telemetry_verify_metrics/tasks/main.yml
+++ b/roles/telemetry_verify_metrics/tasks/main.yml
@@ -78,3 +78,21 @@
     file: verify_openstack_network_exporter_metrics.yml
   tags: test
   when: '"openstack_network_exporter" in telemetry_verify_metrics_metric_sources_to_test'
+
+- name: Check if kepler test VM is running
+  ansible.builtin.shell: |
+    {{ openstack_cmd }} server show {{ kepler_test_vm }} -c status -f value
+  register: kepler_vm_status
+  ignore_errors: true
+  changed_when: false
+  when: '"kepler_vm" in telemetry_verify_metrics_metric_sources_to_test'
+
+- name: Verify Kepler VM metrics are being exposed and stored
+  ansible.builtin.include_tasks:
+    file: verify_kepler_vm_metrics.yml
+  tags:
+    - test
+    - power_monitoring
+  when: '"kepler_vm" in telemetry_verify_metrics_metric_sources_to_test and "ACTIVE" in kepler_vm_status.stdout'
+  vars:
+    metrics_cmd_prefix: "{{ openstack_cmd }} metric show --disable-rbac -c value -f value"
--- a/roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
+++ b/roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
@@ -1,0 +1,33 @@
+- name: |
+    TEST Check that kepler shows idle power consumption metric for VM and it is not zero
+    RHOSO-13583
+  ansible.builtin.shell: |
+    set -euxo pipefail
+    {{ metrics_cmd_prefix }} "kepler_vm_package_joules_total{'vm_id'='{{ kepler_test_vm }}', 'mode'='idle'}"
+  register: result
+  changed_when: false
+  failed_when: result.rc != 0 or result.stdout|float == 0
+
+- name: |
+    TEST Check that kepler shows dynamic metrics for VM and they are increasing
+    RHOSO-13952
+    Step 1 Get an initial value of VM dynamic metrics and check that it's not zero
+  ansible.builtin.shell: |
+    set -euxo pipefail
+    {{ metrics_cmd_prefix }} "kepler_vm_package_joules_total{'vm_id'='{{ kepler_test_vm }}', 'mode'='dynamic'}"
+  register: value1
+  changed_when: false
+  failed_when: value1.rc != 0 or value1.stdout|float == 0
+
+- name: |
+    TEST Check that kepler shows dynamic metrics for VM and they are increasing
+    RHOSO-13952
+    Step 2 Get a new value of VM dynamic metrics and compare with the initial value
+  ansible.builtin.shell: |
+    set -euxo pipefail
+    {{ metrics_cmd_prefix }} "kepler_vm_package_joules_total{'vm_id'='{{ kepler_test_vm }}', 'mode'='dynamic'}"
+  register: value2
+  delay: 10
+  retries: 3
+  changed_when: false
+  until: value2.rc == 0 and (value2.stdout|float > value1.stdout|float)
--- a/ci/vars/osp18_env.yml
+++ b/ci/vars/osp18_env.yml
@@ -6,3 +6,5 @@ stack_flavor: "m1.small"
 stack_network: "private"
 stack_external_network: "public"
 stack_name: "vnf"
+# kepler_test_vm should be created before executing the test suite
+kepler_test_vm: "kepler_test_vm"


roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: |
+    TEST Check kepler shows dynamic (accumulated) power consumption metrics for VM and they are increasing
+    RHOSO-13952
+  ansible.builtin.shell: |
+    set -x
+    params="--disable-rbac kepler_vm_platform_joules_total -c value -c mode -c vm_id -f value"
+    value1=`{{ openstack_cmd }} metric show $params | grep {{ kepler_test_vm }} | grep dynamic | cut -f3 -d ' '`
+    sleep 20
+    value2=`{{ openstack_cmd }} metric show $params | grep {{ kepler_test_vm }} | grep dynamic | cut -f3 -d ' '`
+    (( $(echo "$value2 > $value1" | bc -l) ))
```
elfiesmelfie, line 22: Is it possible to replace the use of bc with some ansible?
Can the two values be converted to floats and then compared?

e.g. something like "(value2 | float) < (value1 | float)"
This could be used in the `failed_when` in the task.

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
```diff
@@ -0,0 +1,27 @@
...
+  register: result
+  delay: 10
```
elfiesmelfie, line 24: I suggest splitting the task, so that there are two separate tasks for getting value1 and value2.

I don't think that the retries here are going to be effective, since you are always getting two values that are 20 second apart. They retries are insurance in case any part of the shell command fails in some other way.
If you split the task, you can get value1 plus wait the 20 seconds, and then retry getting value2 until it changes.

elfiesmelfie (changes requested):

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
```diff
@@ -0,0 +1,27 @@
...
+    params="--disable-rbac kepler_vm_platform_joules_total -c value -c mode -c vm_id -f value"
+    {{ openstack_cmd }} metric show $params | grep {{ kepler_test_vm }} | grep idle | cut -f3 -d ' '
+  register: result
+  delay: 10
+  retries: 3
+  changed_when: false
+  until: result.rc == 0 and result.stdout|float != 0
+
+- name: |
+    TEST Check kepler shows dynamic (accumulated) power consumption metrics for VM and they are increasing
```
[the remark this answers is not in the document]
rsafrono: The metrics I plan to test here are openstack VM metrics that kepler collects. Kepler produces also other metrics but my team is focusing on VM metrics. The tests require a VM pre-spawned. I did it this way in order to avoid pushing workload creation/deletion commands into the tests repo since my team plans to run the tests on NFV baremetal environments and workload creation commands use many NFV specifics. So our CI jobs will handle creation and deletion of NFV-specific ports and servers (I consider adding one more server with different characteristics, e.g. other port types, more CPUs and RAM amount, to compare with the first one, in further tests).
The tests are disabled by default, in case anyone wants to run the tests need to take care of creating the VM with vm_id kepler_test_vm specified in ci/vars/osp18_env.yml and only then enable kepler_vm tests in telemetry_verify_metrics_metric_sources_to_test. I also added a comment in that file that VM should be pre-created.

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
rsafrono: Thanks for the idea. I was thinking that test should be inside a single task. 
I've put now each test step to a separate task. Second one has retries until value increases.

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
rsafrono: done, replaced bc with ansible condition

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
```diff
@@ -0,0 +1,27 @@
+- name: |
+    TEST Check kepler shows idle power consumption metrics for VM and they are not zero
+    RHOSO-13583
+  ansible.builtin.shell: |
+    params="--disable-rbac kepler_vm_platform_joules_total -c value -c mode -c vm_id -f value"
```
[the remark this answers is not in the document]
rsafrono: done, added a var. Unfortunately, not sure how to replace grep. CLI command does not provide proper filtering. Added set -euxo pipefail to ensure task fail in case grep not produces an output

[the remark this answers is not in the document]
vyzigold: You can use the PromQL and specify the correct label values you want for the filtering. PromQL is quite powerful and the observabilityclient supports it fully. See an example of 2 commands here https://paste.opendev.org/show/bxm9HYpIvXLJtW39CQrA/ 

I don't have kepler VM metrics at hand right now to test it, but I guess a one of the commands you want here would look something like: `openstack metric query --disable-rbac -c value -f value "kepler_vm_platform_joules_total{'vm_id'='{{ kepler_test_vm }}', 'mode'='idle'}"`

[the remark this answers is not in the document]
rsafrono: thanks, updated the PR to use this approach

rsafrono:
recheck

rsafrono:
recheck

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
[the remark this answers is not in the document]
elfiesmelfie: If the VM is expected to be created externally, there should, IMHO, be a check in this role to make sure that the VM exists, and skip/fail the tests (with a warning/message) if it doesn't exist.

Something like: Can't run this test, pre-req not met. There is no VM called XXX.
This makes it so much easier to debug if someone naively enables the check for keplar vm metrics.

Alternatively, these tests can be run with the rest of the keplar metrics, if there's a VM present.
So, there's no additional config option to add for baremetal vs VM metrics i.e. check all the keplar metrics you can 

Whatever you choose to do, this dependency should be noted in the README for the role.

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -78,3 +78,13 @@
...
   tags: test
   when: '"openstack_network_exporter" in telemetry_verify_metrics_metric_sources_to_test'
+
+- name: Verify Kepler VM metrics are being exposed and stored
+  ansible.builtin.include_tasks:
+    file: verify_kepler_vm_metrics.yml
+  tags:
+    - test
+    - power_monitoring
+  when: '"kepler_vm" in telemetry_verify_metrics_metric_sources_to_test'
```
elfiesmelfie, line 88: I suggest adding another check for the VM metrics. If `keplar in telemetry_verify_metrics_metric_sources_to_test `, AND the expected VM exists, then check for the VM metrics.

Does it need to be a particular VM, or could you fall back to any VM?

elfiesmelfie (changes requested):

roles/telemetry_verify_metrics/tasks/main.yml
rsafrono: I would prefer checking particular VMs, since I have plans to check also (in some other test) that an extra VM with more CPUs and memory consumes more power.

elfiesmelfie (approved):

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
vyzigold: I have a similar worry as Roman. @elfiesmelfie do you think the XML reporting will be correct with the tasks split like this? The 2 tasks have the same test ID. Unfortunately we can't run it without baremetal machines to see what happens.

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -78,3 +78,20 @@
     file: verify_openstack_network_exporter_metrics.yml
   tags: test
   when: '"openstack_network_exporter" in telemetry_verify_metrics_metric_sources_to_test'
+
+- name: Check if kepler test VM is running
```
vyzigold, line 82: This task needs a "changed_when" according to the linter.

roles/telemetry_verify_metrics/tasks/verify_kepler_vm_metrics.yml
elfiesmelfie: If you're concerned, you can remove the TEST from both, and add an assert task below with the test in in.

e.g. 

Remove "TEST" from the two  tasks
Add a third task that asserts that `value2.rc == 0 and (value2.stdout|float > value1.stdout|float)` (or something similar)

roles/telemetry_verify_metrics/tasks/main.yml
rsafrono: done

vyzigold (approved):

elfiesmelfie (approved):

PR #226, opened 2025-02-19
Add WorkLoad to verify metrics collection

mgirgisf:

PR #240, opened 2025-03-13
[Zuul] Update files_list to be in the project

mgirgisf:
if we have the files listed in the jobs, then any change in telemetry-operator repo will not trigger them.
so I moved the files listed to the project to make the filtering happen in the scope of feature-verification repo.

its being tested in : https://github.com/openstack-k8s-operators/telemetry-operator/pull/644 
and in https://github.com/infrawatch/feature-verification-tests/pull/226

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -110,7 +110,7 @@
     github-check:
       jobs:
         - feature-verification-tests-noop:
-          files: *irrelevant_files
+            files: *irrelevant_files
         - openstack-k8s-operators-content-provider:
             override-checkout: main
             irrelevant-files: *irrelevant_files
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -124,23 +103,46 @@
...
     github-check:
       jobs:
-        - feature-verification-tests-noop
+        - feature-verification-tests-noop:
+            files: *irrelevant_files
         - openstack-k8s-operators-content-provider:
             override-checkout: main
             irrelevant-files: *irrelevant_files
         - functional-tests-on-osp18:
-            voting: true
```
paramite, line 143: Was the voting removal intentional or a typo?

elfiesmelfie: It could be either.

voting is true by default, so removing it should not change behaviour

PR #241, opened 2025-03-17
Add ceilometer prom exporter test

mgirgisf:


mgirgisf:
recheck

mgirgisf:
recheck

mgirgisf:
recheck

mgirgisf:
recheck

roles/telemetry_verify_metrics/tasks/main.yml
```diff
@@ -102,3 +108,4 @@
   when: '"kepler_vm" in telemetry_verify_metrics_metric_sources_to_test and "ACTIVE" in kepler_vm_status.stdout'
   vars:
     metrics_cmd_prefix: "{{ openstack_cmd }} metric show --disable-rbac -c value -f value"
+
```
vyzigold, line 111: this empty line shouldn't be here.

vyzigold:
I have a few questions slightly related to the PR, mostly because I don't remember everything that has happened around the prometheus exporter:

- What about the central-agent prometheus exporter tests? Does it not export metrics through the prometheus exporter? Or would that be tested with a followup PR?
- Also I think the current compute-agent tests could eventually be removed right? But I get it that we want to leave them here for now because of the downstream tests.

ayefimov-1 (approved):

mgirgisf:
recheck

vyzigold (approved):

PR #242, opened 2025-03-18
Remove the check for kepler dashboard

mgirgisf:


roles/telemetry_graphing/files/dashboard-openstack-cloud.js
```diff
@@ -37,8 +37,8 @@ describe('OpenShift Console Dashboard Test', () => {
       { url: '/grafana-dashboard-openstack-node', screenshot: 'openstack-node' },
       { url: '/grafana-dashboard-openstack-vm', screenshot: 'openstack-vms' },
       { url: '/grafana-dashboard-openstack-network-traffic', screenshot: 'openstack-network-traffic'},
-      { url: '/grafana-dashboard-openstack-kepler', screenshot: 'openstack-kepler'},
```
elfiesmelfie, line 40: Is it possible to put these in a separate test somehow? Then we could have these enabled in the powermonitoring job, if we wanted?

I'm not blocking on this, but it would be nice-to-have.

elfiesmelfie (approved):

jlarriba (approved):

PR #249, opened 2025-03-28
[autoscaling] Update initial test for number of instances

elfiesmelfie:
Fail if there's not an instance already created.
Also removes the "wc -l" portion of the shell
commands and uses stdout_lines | length filter


myadla (approved):
looks good

PR #246, opened 2025-03-21
[ci] Update report_results to use XML files

elfiesmelfie:
Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/252

PR #251, opened 2025-04-07
OSPRH-13452: Update functional-tests-on-osp18 to functional-autoscaling-tests-osp18 to match the fvt job name pattern

MiguelCarpio:
Why do we need this PR?

The autoscaling job is not called autoscaling upstream.

The `functional-tests-on-osp18` job should be updated to match the pattern for the other  fvt jobs i.e. `functional-autoscaling-tests-osp18`
 
NOTE: This change affects feature-verification-tests as well as telemetry-operator.

Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/252

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,6 +1,6 @@
 ---
 - job:
-    name: functional-tests-on-osp18
+    name: functional-autoscaling-tests-osp18
     parent: telemetry-operator-multinode-autoscaling
     description: |
       Run the autoscaling functional test on osp18+patched version of aodh and heat.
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
You're hitting the known issues with the graphing test failing.
It is addressed for telemetry-operator in https://github.com/openstack-k8s-operators/telemetry-operator/pull/667

We need a similar change here. The reason you're hitting the issue is that changes to the zuul config trigger the job.
Usually, I'd say "let's make the graphing job non-voting here too!". HOWEVER, the graphing job would typically run for changes to only the relevant test roles, so we don't want it to be non-voting all the time.
I have a dilemma.  I think we'll need to make the graphing temperarily non-voting in order to merge your change and the change at https://github.com/infrawatch/feature-verification-tests/pull/246

elfiesmelfie:
recheck

elfiesmelfie:
Regarding the commit message: because you added an alias job, the telemetry-operator jobs remain unaffected. telemetry-operator will continue to use the old job name, until you make a change there to rename it.

If you want to test to make sure telemetry-operator runs as expected, you can create a PR to T-O which depends-on this one, and see which jobs run

MiguelCarpio:
> Take care, you need to make change to the downstream also before remove the alias.

Thanks @mgirgisf, I've added that note to https://issues.redhat.com/browse/OSPRH-13452

elfiesmelfie (approved):

vyzigold (approved):

PR #255, opened 2025-04-11
Set default values for *_test_id vars

elfiesmelfie:
The *test_id var was set for the common and telemetry_logging roles previously, so that the tests would be reported as passed/failed by custom_logger

Since report_result was updated to use XML files to determine failure, the test ID does not need to be set.
This change adds a default value for these vars, so that we can start removing the vars from test configs.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -1,2 +1,14 @@
 ---
 openstack_cmd: "openstack"
+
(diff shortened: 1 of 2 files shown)


elfiesmelfie:
recheck

vyzigold (approved):

mgirgisf (approved):

PR #256, opened 2025-04-14
[ci] Try to combine the FVT and tempest

elfiesmelfie:
This approach runs the FVTs before tempets and then collects the results after running test-operator This means that the tempest tests will trigger a failure if they fail, and the FVTs will trigger a failure only after the tempest tests are run, which means that the results are reported as expected.

This will work for github-check pipeline and the component pipeline. However, this approach may not be usable with the unified jobs.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -110,7 +110,6 @@
     description: |
       Run the UI Graphing test and tempest smoketests on osp18.
     vars:
-      cifmw_run_tests: false
       cifmw_extras:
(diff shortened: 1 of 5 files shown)


elfiesmelfie:
recheck

elfiesmelfie:
I still need to do the following:
* enable the autoscaling tempest test
* Move FVT results to tests/ instead of logs
* Incorporate other WIP changes to the jobs

lnatapov (commented):
LGTM

PR #266, opened 2025-04-29
[zuul] Re-parent the verify-metrics job

elfiesmelfie:
It runs the same tests as autoscaling-fvt, so can become an alias.

In the future, it may select a subset of tests to run via some var, but for now is the same as the autoscaling-fvt job


vyzigold (approved):

ayefimov-1 (approved):
looks good.

PR #269, opened 2025-05-22
Fix verify metrics rabbitmq command

vyzigold:
The command was missing the --disable-rbac option, which caused the test to fail after update to python-observabilityclient 0.5.0.


elfiesmelfie:
recheck

PR #272, opened 2025-05-30
[ci] Use pre_tests hooks for running FVTs

elfiesmelfie:
In some cases we were using post_deploy, and others pre_test for running the FVTs. These happen sequentially, but for updates jobs, the post_deploy hooks are not run. In order for the tests to be run in updates jobs, the pre_tests hook should be used.


vyzigold (approved):
Not sure what's happening with the request for review vs. the PR being a draft, but the change looks good and the CI seems to be passing.

PR #273, opened 2025-05-30
[zuul] Add logging config and tests to functional-autoscaling-tests-osp18

elfiesmelfie:


vyzigold:
Found this while doing my round as CI watcher. Interesting. I wonder what's exactly causing the job to take so long. I took a quick look at the must-gather and it looks like logging is enabled and it might even work as far as I can tell. But autoscaling and metricstorage is disabled now. Looking at https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/ci/vars-logging.yml and https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/ci/vars-autoscaling.yml , both define the `cifmw_edpm_prepare_kustomizations` variable, which basically causes autoscaling to never get enabled. The kustomizations will either need to manually merge or maybe there is some ci-framework magic, which could take 2 kustomizations and apply both of them.

elfiesmelfie:
> Found this while doing my round as CI watcher. Interesting. I wonder what's exactly causing the job to take so long. I took a quick look at the must-gather and it looks like logging is enabled and it might even work as far as I can tell. But autoscaling and metricstorage is disabled now. Looking at https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/ci/vars-logging.yml and https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/ci/vars-autoscaling.yml , both define the `cifmw_edpm_prepare_kustomizations` variable, which basically causes autoscaling to never get enabled. The kustomizations will either need to manually merge or maybe there is some ci-framework magic, which could take 2 kustomizations and apply both of them.

The main issue is that the logging vars have changed, but the logging job has not been updated to deploy metrics storage or autoscaling. This will eventually be addressed in https://github.com/openstack-k8s-operators/telemetry-operator/pull/698.
This PR was the spike part of the work, where I hack it together with whatever workarounds are needed in order to identify the effort needed to do it properly.

elfiesmelfie:
recheck

vyzigold (approved):

elfiesmelfie:
recheck

mgirgisf (approved):

.zuul.yaml
```diff
@@ -32,6 +35,7 @@
         override-checkout: main
       - name: github.com/openstack-k8s-operators/telemetry-operator
         override-checkout: main
+    timeout: 18000  # 5 hours
```
vyzigold, line 38: Is this planed to stay? Do we expect some significant extension of how long the jobs run?

elfiesmelfie: No, I think it can be removed.
There were some timeouts initially, caused by autoscaling not being deployed, so I extended the timeout to get some results and errors to help debug.

elfiesmelfie:
recheck

vyzigold (approved):

PR #275, opened 2025-06-02
OSPRH-16753: Converge autoscaling logging & graphing tests based on enabled features

MiguelCarpio:
Converge the following tests based on the OpenStackControlPlane telemetry enabled features:

- logging
- autoscaling
- graphing

`functional-logging-tests-osp18`, `functional-autoscaling-tests-osp18` and `functional-graphing-tests-osp18` will be replaced by `functional-tests-osp18` job, which runs `var_functional-tests.yml` that plays the `run_functional_tests.yml` playbook. 

`run_functional_tests.yml` runs the following tests:

- run_logging_tests_computes.yml
- run_logging_tests_controller.yml
- [1] run_autoscaling_tests.yml
- [1] run_graphing_test.yml

[1] If the `run_autoscaling_test` and `run_graphing_test` extra_vars are true, and if the `autoscaling` and/or the `dashboardsEnabled` features are enabled.


ci/run_cloudops_tests_osp18.yml
elfiesmelfie, line 1: Same feedback here for the pre_tasks. It will reduce the size of the change, and reduce the possibility of errors later, if new tasks are needed for the autoscaling tests, for example.

ci/run_cloudops_tests_osp18.yml
```diff
@@ -13,10 +13,12 @@
         - name: Include vars from the extra_vars files
           ansible.builtin.include_vars:
             dir: "{{ cifmw_basedir }}/artifacts/parameters"
+          when: hostvars['localhost'].enable_autoscaling_role | bool
```
elfiesmelfie, line 16: Guarding the block is sufficient here. The individual task don't need to be guarded.

ci/run_fvt_tests.yml
elfiesmelfie, line 1: If going with the approach of using pre_tasks (as suggested above), then the check for each set of tests could happen in the pre_task section of the play.

There's an alternative approach where the tests are in one large playbook, with multiple plays that will detect whether the particular features are enabled, then try to run all the tests, with guarding in the appropriate place.

MiguelCarpio:
recheck

MiguelCarpio:
recheck

.zuul.yaml
```diff
@@ -58,13 +59,29 @@
...
+    parent: functional-autoscaling-graphing-tests-osp18
     extra-vars: *functional_autoscaling_extra_vars
     description: |
-      functional-tests-on-osp18 is an alias of functional-autoscaling-tests-osp18, 
+      functional-tests-on-osp18 is an alias of functional-autoscaling-graphing-tests-osp18, 
       temporary added until openstack-k8s-operators/telemetry-operator updates 
       references of functional-tests-on-osp18 jobs.
 
+- job:
+    name: functional-autoscaling-tests-osp18
```
MiguelCarpio, line 70: I need the `functional-autoscaling-tests-osp18` and `functional-graphing-tests-osp18` aliases, which aim to `functional-autoscaling-graphing-tests-osp18`, to allow the telemetry operator and GitHub to validate this PR.

.zuul.yaml
```diff
@@ -159,10 +157,18 @@
         - functional-graphing-tests-osp18:
```
MiguelCarpio, line 157: Once this PR is approved, the `functional-graphing-tests-osp18` and functional-autoscaling-tests-osp18 should be replaced by `functional-autoscaling-graphing-tests-osp18`. This PR handle it https://github.com/openstack-k8s-operators/telemetry-operator/pull/717

.zuul.yaml
```diff
@@ -1,10 +1,10 @@
 ---
 - job:
-    name: functional-autoscaling-tests-osp18
+    name: functional-autoscaling-graphing-tests-osp18
```
elfiesmelfie, line 3: IMHO, there is no need to rename this job because of the extra tests being added. 
This removes the need to edit telemetry-operator to update the job name

elfiesmelfie (changes requested):

.zuul.yaml
vyzigold: Do you actually need a separate PR for that? I'm pretty sure you can just do all the changes here. Also I think if you delete the graphing job definition in this PR (which happens a little above this), then the FVT github-check will be broken until the follow-up change merges.

vyzigold:
recheck - the last failure didn't seem related to the PR

vyzigold:
So before, it seemed like the whole crc failed somehow. We didn't even have must-gather logs as it just seemed it couldn't access the cluster. On the re-run, FVT tests are successful, but tempest didn't run, because it can't connect to keystone. In keystone logs, I see issues with galera and in galera logs, I see a bunch of aborted connections and then it restarted once. Based on the timestamps, it looks like it all happened around the same time. Let's try once more.

.zuul.yaml
```diff
@@ -137,6 +119,7 @@
     vars:
       patch_observabilityclient: true
       cifmw_update_containers: false
+      crc_enable_monitoring: true
```
vyzigold, line 122: Regarding the failures you're seeing with this job. I took a look at the events in the must gather https://logserver.rdoproject.org/ec1/rdoproject.org/ec1a0aaeaefd4e52a190e144fc50b6e8/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/events.log and what I see there is, that the tempest pod starts and all the probes on a lot of other pods start failing (including Prometheus from our metric-storage, which isn't connected to OpenStack as much as the rest of the pods). This causes a lot of restarts, services are unavailable and tempest tests fail.

What I think could be happening is, that enabling openshift monitoring is probably adding a lot of additional load to an already resource intensive deployment (in comparison with our other jobs, we're running ceph here, which I guess eats quite a few resources as well). So my proposal is, let's just not run graphing tests as part of this job (its main purpose is to test volume pool metrics, the rest of the tests are there because why not). So, can you remove this line and somehow set ` play_graphing_test: true`, so that the graphing tests don't run? I'm not sure where that variable would need to go in order to take effect though.

vyzigold: If that doesn't help, then I don't have any more ideas at the moment, so my next proposal would be to set it non-voting and I'll get back to it later.

MiguelCarpio:
recheck

vyzigold (approved):

PR #278, opened 2025-06-09
Remove RHOSO-XXX IDs from the FVTs

myadla:
Remove all RHOSO-XXX IDs from the FVTs and replace them with "TEST"
The RHOSO-XXX IDs are removed, since all RHOSO tests use XML for reporting to polarion


roles/common/tasks/container_test.yml
```diff
@@ -24,8 +24,7 @@
       var: debug_output.stdout_lines
 
 - name: |
-    TEST Verify {{ container_name }} container status
-    {{ common_container_test_id }}
+    {{ common_container_test_id }} Verify {{ container_name }} container status
```
elfiesmelfie, line 27: The "TEST" prefix should stay where it is, it does not need to be updated in this way.

Removing the IDs from the tasks should involve removing the usage of the `common_XXX_test_id` vars  from the test roles and from the test configurations in ci/

elfiesmelfie (changes requested):

roles/common/tasks/container_test.yml
myadla: Made the required changes

ci/report_result.yml
elfiesmelfie, line 1: restore this file

callback_plugins/custom_junit.py
elfiesmelfie, line 1: Restore this file

ci/vars/common.yml
elfiesmelfie, line 1: restore this file

elfiesmelfie (changes requested):
Restore the files that were removed.

Update the roles/common/README to remove mention of these test_id vars.
Update roles/common/main.yml to remove the uses of the test_id vars

PR #279, opened 2025-06-09
[common] Update manifest role and config

elfiesmelfie:
* Use a list of strings, instead of a space-separated string
* Remove the string separating tasks
* Replace `wc -l` with stdout_lines | length
* Name the loop var

### DIFF
--- a/ci/logging_tests_controller.yml
+++ b/ci/logging_tests_controller.yml
@@ -46,9 +46,8 @@
 
     common_manifest_test_id: "RHOSO-12677"
     common_manifest_list:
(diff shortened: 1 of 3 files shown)


ayefimov-1 (dismissed):
That looks good.

vyzigold (commented):
Maybe instead of checking the number of package manifests, we could check the catalog? e.g. we care for the loki-operator from the redhat-operators catalogSource. The community loki-operator can come and go and we shouldn't really care in our tests, since we don't use it anywhere.

PR #282, opened 2025-06-12
[ci] Update the logging tests to use new pod test config format

elfiesmelfie:
Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/281


elfiesmelfie:
recheck

PR #283, opened 2025-06-13
[zuul] Add roles/common to the autoscaling and logging jobs

elfiesmelfie:
The metrics verification tests in the autoscaling job use the common/ role The logging jobs use the common role.
Changes to roles/common should trigger both of these jobs to run.


ayefimov-1 (approved):
Looks good!

PR #291, opened 2025-06-25
Remove unneeded test IDs from FVTs

myadla:
Removed all rhosoid's from roles/telemetry_autoscaling/tasks


myadla:
recheck

elfiesmelfie (approved):

myadla:
recheck

mgirgisf:
the problem here is not related to the jobs them self, but in  report_result hook ,  https://github.com/infrawatch/feature-verification-tests/blob/207992e2b0b213081d5c7e2c688e523bbdece275/ci/report_result.yml#L43 

As when we remove all the IDs the file "test_run_result.out" will not be created and then the hook fail when we collect the result. so maybe we should  update the logger callback plugin to create the file even if its empty, or maybe better is to remove using this file if we don't need it anymore .

PR #292, opened 2025-06-26
Deploy monitoring and cluster-monitoring-operator in CRC

danpawlik:
The "observe" submenu in OpenShift console is required for telemetry team for some tests.
Enable monitoring and CMO feature in CRC.


danpawlik:
cc @elfiesmelfie @mgirgisf

.zuul.yaml
```diff
@@ -96,6 +96,7 @@
     description: |
       Run the UI Graphing test and tempest smoketests on osp18.
     vars:
+      crc_enable_monitoring: true
```
elfiesmelfie, line 99: Wait, does crc_enable_monitoring work with extra-vars?
If yes, then this should eventually be migrated into the graphing deployment vars file in https://github.com/openstack-k8s-operators/telemetry-operator/pull/696 (CC: @MiguelCarpio )
If no, then this is fine as-is and needs no updates.

danpawlik: TBH don't know :confused:

PR #298, opened 2025-07-04
[ci] Remove the extra copy of XML file from hooks

elfiesmelfie:
The logs are output to the expected location, and don't need to be copied twice.


elfiesmelfie:
@mgirgisf I've set this PR to automerge, but it still needs 2 approvals from code review

PR #303, opened 2025-07-18
Modify test id's with "rhelosp-#####" to TEST under feature-verification-tests/roles/

myadla:
Modify test id's with "rhelosp-#####" to TEST under feature-verification-tests/roles/test-<alerts>


ayefimov-1 (approved):
looks good.

PR #307, opened 2025-08-05
Add disable-rbac to volume metric tests

vyzigold:
The volume metrics don't have a project label, so older versions of the python-observabilityclient aren't able to show the metrics unless we use the --disable-rbac.


vyzigold:
recheck

elfiesmelfie (approved):

mgirgisf (approved):

PR #312, opened 2025-09-15
Add test for the PrometheusEndpoint secret

mgirgisf:


mgirgisf:
> While the test itself seems pretty good to me. I don't think it fits into the telemetry_verify_metrics very well as this doesn't test any metrics or anything related to some metric source.
> 
> I think the default-telemetry would be a better place for this. default-telemetry has scenarios for testing how metric-storage and autoscaling behave under different circumstances and it checks various resources created by metric-storage. Victoria is actually adding NAD tests to default-telemetry, which is kind of a similar category of test to this.

Thanks Jaromir .. I had my doubt also about the best place to add this test , so let me close this and added to the default telemetry

PR #334, opened 2026-02-02
Add task files to push and retrieve ck data log to loki

myadla:
Task files below takes up synthetic json data file from Cloudkitty and pushes the data, flush the data to loki server and queries data back from loki
1. setup_loki_env.yml
2. ingest_loki_data.yml
3. flush_loki_data.yml
4. retrieve_loki_data.yml


myadla:
recheck

roles/telemetry_chargeback/README.md
```diff
@@ -62,6 +66,25 @@ Dependencies
 ------------
 This role has no direct hard dependencies on other Ansible roles.
 
+This runs 5 playbooks
```
elfiesmelfie: These are not playbooks, they are tasks in the role.

roles/telemetry_chargeback/README.md
```diff
@@ -78,4 +101,4 @@ Example Playbook
 Author Information
 ------------------
 
-Alex Yefimov, Red Hat
+Alex Yefimov, Muneesha Yadla
```
elfiesmelfie:
```suggestion
Alex Yefimov, Red Hat
Muneesha Yadla, Red Hat
```

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,79 @@
...
+- name: Extract Client Certificates
+  ansible.builtin.command:
+    cmd: "oc extract {{ client_secret }} --to={{ local_cert_dir }} --confirm -n {{ namespace }}"
+  changed_when: true
+
+# Extract CA Bundle
+- name: Extract CA Bundle
+  ansible.builtin.command:
+    cmd: "oc extract {{ ca_configmap }} --to={{ local_cert_dir }} --confirm -n {{ namespace }}"
+  changed_when: true
```
elfiesmelfie: This could be refactored into a separate task file, to keep the focus on the flush_loki_data tasks.

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,79 @@
...
+      changed_when: true
+      failed_when: flush_response.rc != 0
+
+    # Status
+    - name: Verify Flush Status
+      ansible.builtin.assert:
+        that:
+          - "'204' in flush_response.stderr or '200' in flush_response.stderr"
+        fail_msg: "Flush failed"
+        success_msg: "Ingester Memory Flushed successfully"
```
elfiesmelfie, line 37: Nice! This usage of assert to check the status and then give a message either way makes it easy to debug what's happening.

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    - name: Confirm Success
+      ansible.builtin.debug:
+        msg: "Loki Ingester flushed the data successfully"
```
elfiesmelfie:
```suggestion
```

This debug message is unnecessary since the assert task already prints the success message

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Debug Failure Output
+      ansible.builtin.debug:
+        msg: 
+          - "Failure"
+          - "Stdout: {{ flush_response.stdout }}"
+          - "Stderr: {{ flush_response.stderr }}"
+    
+    - name: Report Failure
+      ansible.builtin.fail:
+        msg: "Flush failure"
```
elfiesmelfie:
```suggestion
```

This is not necessary, since the failures is reported in the previous debug task

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,79 @@
...
+
+# Cleanup
+- name: Cleanup local certificates
+  ansible.builtin.file:
+    path: "{{ local_cert_dir }}"
+    state: absent
```
elfiesmelfie:
```suggestion
  always:
      # Cleanup
      - name: Cleanup local certificates
        ansible.builtin.file:
          path: "{{ local_cert_dir }}"
          state: absent
```

Using the always block ensures that the certificated are removed even when there's a failure.

roles/telemetry_chargeback/tasks/ingest_loki_data.yml
```diff
@@ -0,0 +1,59 @@
...
+
+# Extract certificates
+- name: Extract certificates from openshift secret
+  ansible.builtin.command:
+    cmd: >
+      oc extract secret/{{ cert_secret_name }}
+      --to={{ cert_dir }}
+      --confirm
+  changed_when: true
+
```
elfiesmelfie: This can be combined into the suggest prepare_certs task file for flush_loki_data.

The tasks/prepare_certs can deal with extracting the certs and putting them into the right places.

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -7,3 +7,32 @@ ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
...
+loki_base_url: "https://cloudkitty-lokistack-openstack.apps-crc.testing"
+loki_push_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/push"
+loki_query_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/query_range"
+
+# Ingester URL
+ingester_flush_url: "https://cloudkitty-lokistack-ingester-http.openstack.svc:3100/flush"
+
+# Cloudkitty certificates
+cert_secret_name: "cert-cloudkitty-client-internal"
+cert_dir: "{{ ansible_user_dir }}/ck-certs"
```
elfiesmelfie: If the certs are always removed, it is fine to define the vars here.

However, if the certs are ever kept around for debugging, and you want to be able to modify the value of cert_dir, then this should be added to defaults/main instead.

Vars defined in the vars/ dir have higher precedence that play vars_file, so we cannot override these the way we do for other roles.

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -7,3 +7,32 @@ ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
...
+# loki url's
+loki_base_url: "https://cloudkitty-lokistack-openstack.apps-crc.testing"
+loki_push_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/push"
+loki_query_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/query_range"
+
+# Ingester URL
+ingester_flush_url: "https://cloudkitty-lokistack-ingester-http.openstack.svc:3100/flush"
+
+# Cloudkitty certificates
+cert_secret_name: "cert-cloudkitty-client-internal"
```
elfiesmelfie: This should be moved to defaults/main (see comment on line 22), as it is likely that we could need to override it.

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -7,3 +7,32 @@ ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
...
+ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
+local_cert_dir: "{{ ansible_env.HOME }}/flush_certs"
+remote_cert_dir: "osp-certs"
+
+# LogQL Query
+logql_query: "{{ loki_query | default('{service=\"cloudkitty\"}') }}"
+
+# Time window settings
+lookback_days: "{{ lookback | default(6) }}"
+limit_logs: "{{ limit | default(50) }}"
```
elfiesmelfie: Move these to defaults/main, and you can then set the defaults there directly, and remove the extra `lookback` and `limit` vars.

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -7,3 +7,32 @@ ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
 ck_data_config: "{{ role_path }}/files/test_static.yml"
 ck_output_file_local: "{{ artifacts_dir_zuul }}/loki_synth_data.json"
 ck_output_file_remote: "{{ logs_dir_zuul }}/gen_loki_synth_data.log"
+ck_loki_retrieve_file: "{{ logs_dir_zuul }}/retrieve_loki_op.json"
+
+# loki url's
+loki_base_url: "https://cloudkitty-lokistack-openstack.apps-crc.testing"
```
elfiesmelfie: This should be in defaults/main, so that it can be reconfigured.

This is currently tightly coupled to a CRC-based deployment. This won't work on a non-CRC OCP deployment.
So this can't be run in any uni-job testing or customer-like environments, since they will have other URLs.


I believe that we can assume the existance of certain routes in the deployment, and use those to get the values for the loki_base_url and loki_ingester_url, in some set up tasks. @vyzigold, does the telemetry-operator (or lokistack) create a set of routes for loki services?

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -7,3 +7,32 @@ ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
...
+# Ingester URL
+ingester_flush_url: "https://cloudkitty-lokistack-ingester-http.openstack.svc:3100/flush"
+
+# Cloudkitty certificates
+cert_secret_name: "cert-cloudkitty-client-internal"
+cert_dir: "{{ ansible_user_dir }}/ck-certs"
+
+client_secret: "secret/cloudkitty-lokistack-gateway-client-http"
+ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
+local_cert_dir: "{{ ansible_env.HOME }}/flush_certs"
```
elfiesmelfie: Are the flush certs different from the local ones?

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
```diff
@@ -0,0 +1,92 @@
...
+          - loki_response.json.data.result | length > 0
+        fail_msg: "Query did not return any data entries"
+        success_msg: "Query found all the data entries"
+
+    # Success
+    - name: Confirm Success
+      vars:
+        actual_count: "{{ loki_response.json.data.result | map(attribute='values') | map('length') | sum }}"
+      ansible.builtin.debug:
+        msg: "Successful, Input file had {{ expected_log_count }} entries and Loki returned {{ actual_count }}"
```
elfiesmelfie: Make this an assert instead, in case the counts are different for some reason.

elfiesmelfie (changes requested):

roles/telemetry_chargeback/vars/main.yml
vyzigold: Yes.

With RedHat loki-operator, there is a observatorium installed in front of loki. So most of our requests go through observatorium (that uses the "local" certs). But it seems like observatorium doesn't expose the "flush" endpoint. So the tasks here bypass the observatorium and tap directly into loki.

roles/telemetry_chargeback/vars/main.yml
vyzigold: We could use this to retreive the route:
```
oc get route cloudkitty-lokistack -ojsonpath --template "{.status.ingress[].host}"
```

roles/telemetry_chargeback/tasks/flush_loki_data.yml
myadla: Thank you

roles/telemetry_chargeback/tasks/flush_loki_data.yml
myadla: Here we need this task for flush itself and not required anywhere else. Also, the same kind (extract certs) are present for ingest_loki.yaml as well but with different certs and gateway URL (and are only 1 time usage)

roles/telemetry_chargeback/vars/main.yml
myadla: working on it

roles/telemetry_chargeback/tasks/flush_loki_data.yml
myadla: working on it

roles/telemetry_chargeback/vars/main.yml
myadla: Added task to verify the route instead of hard coded value for loki push/query.
Need to verify if there is a way to find Ingester URL to flush the data

roles/telemetry_chargeback/tasks/flush_loki_data.yml
myadla: Added all the setup tasks (loki gateway URL's & extracting tasks) in a seperate task file setup_loki_env.yml

myadla:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,28 @@
...
+# LogQL Query
+logql_query: "{{ loki_query | default('{service=\"cloudkitty\"}') }}"
+
+# vars
+namespace: "openstack"
+openstackpod: "openstackclient"
+
+# Time window settings
+lookback: 6
+limit: 50
```
vyzigold: A bit pedantic maybe, but lets fix the file ends.

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,28 @@
...
 openstack_cmd: "openstack"
+
+# Cloudkitty certificates
+cert_secret_name: "cert-cloudkitty-client-internal"
+cert_dir: "{{ ansible_user_dir }}/ck-certs"
+
+client_secret: "secret/cloudkitty-lokistack-gateway-client-http"
+ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
+remote_cert_dir: "osp-certs"
+local_cert_dir: "{{ ansible_env.HOME }}/flush_certs"
```
vyzigold: The ansible is accessing 2 endpoints in a similar manner, each of which is mTLS protected and need a client certificate / key. From these variable names, it's a little confusing what they mean (I can decipher the meaning from the values and from having the context freshly in my mind, but I'll not be able to do that in a year). Let's mark what's what in the var names. WDYT about:

```suggestion
gateway_client_cert_secret_name: "cert-cloudkitty-client-internal"
gateway_client_cert_dir: "{{ ansible_user_dir }}/ck-certs"

ingester_client_cert_secret: "secret/cloudkitty-lokistack-gateway-client-http"
ingester_ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
ingester_client_cert_dir: "{{ ansible_env.HOME }}/flush_certs"

remote_cert_dir: "osp-certs"
```

vyzigold: Also I think `ansible_env.HOME` and `ansible_user_dir` should be the same, so I think that should be unified at least across this file.

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,28 @@
...
+
+
+# Loki Ingest & Retrieve logs
+logs_dir_zuul: "/home/zuul/ci-framework-data/logs"
```
vyzigold: IMO using ansible_user_dir or ansible_env.HOME here would make local execution easier (one less var to required to pass)

roles/telemetry_chargeback/defaults/main.yml
vyzigold: in other files as well

roles/telemetry_chargeback/tasks/main.yml
```diff
@@ -1,6 +1,30 @@
...
+  ansible.builtin.include_tasks: 
+    file: setup_loki_env.yml
+  tags: setup
+
+- name: Ingests Cloudkitty Data log
+  ansible.builtin.include_tasks: 
+    file: ingest_loki_data.yml
+  tags: ingest
+
+- name: Flush Data to loki Storage
```
vyzigold:
```suggestion
- name: Flush Data to Loki Storage
```
To unify the name with how it's capitalized on L12

roles/telemetry_chargeback/tasks/main.yml
```diff
@@ -1,6 +1,30 @@
...
+  ansible.builtin.include_tasks: 
+    file: flush_loki_data.yml
+  tags: flush
+
+- name: Retrieve Data log from loki
```
vyzigold:
```suggestion
- name: Retrieve Data log from Loki
```
To unify the name with how it's capitalized on L12

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
```diff
@@ -0,0 +1,88 @@
...
+      - "Query: {{ logql_query }}"
+      - "Start Time: {{ start_time }}"
+      - "Limit: {{ limit }}"
+
+# Query Loki
+- name: Retrieve Logs from Loki via API
+  block:
+    - name: Query Loki API
+      ansible.builtin.uri:
+        url: "{{ loki_query_url }}?query={{ logql_query | urlencode }}&start={{ start_time }}&limit={{ limit }}"
```
vyzigold: Question... Do we need the limit? What happens if it isn't there? Reason I'm asking, is that if ever somebody changes the values in https://github.com/infrawatch/feature-verification-tests/blob/master/roles/telemetry_chargeback/files/test_static.yml#L4C1-L6C21 Then I think more than 50 values would get generated (and ingested) and the limit here would need to be adjusted as well. So IMO if we can do without a limit, we could save us a headache in the future.

roles/telemetry_chargeback/defaults/main.yml
myadla: Done

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
myadla: Sure..

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -1,9 +1,7 @@
 ---
-logs_dir_zuul: "/home/zuul/ci-framework-data/logs"
 artifacts_dir_zuul: "/home/zuul/ci-framework-data/artifacts"
 
 ck_synth_script: "{{ role_path }}/files/gen_synth_loki_data.py"
 ck_data_template: "{{ role_path }}/template/loki_data_templ.j2"
 ck_data_config: "{{ role_path }}/files/test_static.yml"
-ck_output_file_local: "{{ artifacts_dir_zuul }}/loki_synth_data.json"
-ck_output_file_remote: "{{ logs_dir_zuul }}/gen_loki_synth_data.log"
+ck_output_file_local: "{{ artifacts_dir_zuul }}/loki_synth_data.json"
```
vyzigold: one more file ending

vyzigold (changes requested):

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
```diff
@@ -0,0 +1,87 @@
...
+        body_format: json
+      register: loki_response
+      # Wait condition
+      until:
+        - loki_response.status == 200
+        - loki_response.json.status == 'success'
+        - loki_response.json.data.result | length > 0
+        - (loki_response.json.data.result | map(attribute='values') | map('length') | sum) >= expected_log_count|int
+      retries: 10
+      delay: 150
```
elfiesmelfie: The delay is too long IMHO. The values should be tuned for more retries, and smaller delay.

ayefimov-1 (dismissed):
looks fine.

elfiesmelfie (commented):
I added some comment on this PR, but most of the review is on https://github.com/infrawatch/feature-verification-tests/pull/335, which contains some of this content as well, IINM

vyzigold (dismissed):

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -0,0 +1,70 @@
...
+  ansible.builtin.file:
+    path: "{{ ingester_client_cert_dir }}"
+    state: directory
+    mode: '0755'
+
+- name: Extract Client Certificates
+  ansible.builtin.command:
+    cmd: > 
+      oc extract {{ ingester_client_cert_secret }}
+      --to={{ ingester_client_cert_dir }}
```
elfiesmelfie, line 59: This could use a temp file instead of setting a var in the role.
This has been done in the autoscaling role, IINM, for the files that we copy into the openstackclient pod

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
myadla: In my testing, I have observed that it takes a min of ~15 mins to flush all the data

myadla: Modified to wait for 60 secs

roles/telemetry_chargeback/tasks/setup_loki_env.yml
myadla: Here Ansible is accessing 2 different URL's and client certificates in a similar manner, each of which is mTLS protected and need a client certificate / key. 
So, in order to differentiate them for better understanding(in future too), we are using the var names specifically.

**Loki URL uses these certs to ingest/retrieve:**
gateway_client_cert_secret_name: "cert-cloudkitty-client-internal"
gateway_client_cert_dir: "{{ ansible_user_dir }}/ck-certs"

**Ingester flush URL uses these certs to flush:**
ingester_client_cert_secret: "secret/cloudkitty-lokistack-gateway-client-http"
ingester_ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
ingester_client_cert_dir: "{{ ansible_user_dir }}/flush_certs"

myadla:
@elfiesmelfie Addressed all my changes as per the comments for PR#334.. Please check

myadla:
> I added some comment on this PR, but most of the review is on #335, which contains some of this content as well, IINM

I have addressed the needed for PR#334

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,25 @@
...
+gateway_client_cert_secret_name: "cert-cloudkitty-client-internal"
+gateway_client_cert_dir: "{{ ansible_user_dir }}/ck-certs"
+
+ingester_client_cert_secret: "secret/cloudkitty-lokistack-gateway-client-http"
+ingester_ca_configmap: "cm/cloudkitty-lokistack-ca-bundle"
+ingester_client_cert_dir: "{{ ansible_user_dir }}/flush_certs"
+osp_remote_cert_dir: "osp-certs"
+
+# Loki Ingest & Retrieve logs
+ck_output_file_remote: "{{ logs_dir }}/gen_loki_synth_data.log"
```
elfiesmelfie, line 14: This var gets removed in the next PR (#335 ), so should not be in this PR either. 
This is so that we're minimising the changes that are added and then immediately removed

myadla: As discussed in slack, @ayefimov-1 has incorporated all files from PR#334 into PR#335 and made changes in vars/main.yml to make the them dynamic..

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
vyzigold: 15 minutes??? That's too long. If it takes this long, then imo the flush isn't really doing anything.

myadla: Yes.. I have tested to retrieve the same data without flush, and it takes 60 to 90 mins.. So flush is actually working but its not instant..

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
```diff
@@ -0,0 +1,87 @@
...
+        actual_count: "{{ loki_response.json.data.result | map(attribute='values') | map('length') | sum }}"
+      ansible.builtin.assert:
+        that:
+          - loki_response.json.status == 'success'
+          - loki_response.json.data.result | length > 0
+          - actual_count|int == expected_log_count|int
+        fail_msg: "Query did not return all data entries. Expected {{ expected_log_count }} log entries, but Loki only returned {{ actual_count }}"
+        success_msg: "Query returned all data entries. Input file had {{ expected_log_count }} entries and Loki returned {{ actual_count }}"
+
+  rescue:
```
vyzigold, line 77: The rescue block seems unreachable. See the following output from the logs:

```
2026-02-14 19:34:01,174 p=37156 u=zuul n=ansible | TASK [telemetry_chargeback : Verify Data Integrity] ****************************
2026-02-14 19:34:01,174 p=37156 u=zuul n=ansible | Saturday 14 February 2026  19:34:01 +0000 (0:00:00.382)       0:08:12.788 ***** 
2026-02-14 19:34:01,207 p=37156 u=zuul n=ansible | fatal: [localhost]: FAILED! => {
    "assertion": "actual_count|int == expected_log_count|int",
    "changed": false,
    "evaluated_to": false,
    "msg": "Query did not return all data entries. Expected 24 log entries, but Loki only returned 46"
}
2026-02-14 19:34:01,207 p=37156 u=zuul n=ansible | ...ignoring
2026-02-14 19:34:01,215 p=37156 u=zuul n=ansible | TASK [telemetry_chargeback : Get total rate from loki: test_static_basic_10x] ***
2026-02-14 19:34:01,215 p=37156 u=zuul n=ansible | Saturday 14 February 2026  19:34:01 +0000 (0:00:00.040)       0:08:12.829 ***** 
2026-02-14 19:34:01,234 p=37156 u=zuul n=ansible | included: /home/zuul/src/github.com/infrawatch/feature-verification-tests/roles/telemetry_chargeback/tasks/loki_rate.yml for localhost
```

The assert above fails (it prints the fail_msg), but the rescue block doesn't execute.

myadla: It looks like "ignore_errors:true" is being inherited from ci. I have now explicitly added to override any inherited settings by adding ignore_errors: false in the tasks

myadla:
recheck

myadla:
recheck

myadla:
recheck

myadla:
Addressed all the request changes and rerun completed

myadla:
recheck

myadla:
recheck

myadla:
recheck

myadla:
recheck

myadla:
recheck

myadla:
recheck

myadla:
Made changes according to Jira https://redhat.atlassian.net/browse/OSPRH-27873 (Retrieve Logs from Loki via API {{ item }}" task needs to be enhanced)

myadla:
recheck

PR #337, opened 2026-02-18
[telemetry_verify_metrics] Add node exporter filesystem verification …

jlarriba:
…tests

Verify that node_exporter reports host filesystem metrics instead of container-internal ones, following the --path.rootfs=/rootfs change in edpm-ansible PR #1119.

### DIFF
--- a/roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
+++ b/roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
@@ -1,3 +1,4 @@
+---
 - name: Verify node exporter scrapeconfigs exist
   ansible.builtin.include_role:
     name: common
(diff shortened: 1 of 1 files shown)


roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
```diff
@@ -36,4 +37,34 @@
...
+  register: result
+  delay: 30
+  retries: 10
+  changed_when: false
+  until: result.rc == 0 and "mountpoint" in result.stdout
+
+- name: |
+    TEST Verify node exporter filesystem metrics report a real filesystem type instead of overlay
+  ansible.builtin.shell: |
+    {{ openstack_cmd }} metric query --disable-rbac 'node_filesystem_size_bytes{mountpoint="/"}' | grep fstype
```
elfiesmelfie:
```suggestion
  ansible.builtin.command: |
    {{ openstack_cmd }} metric query --disable-rbac 'node_filesystem_size_bytes{mountpoint="/"}'
```
```suggestion
    {{ openstack_cmd }} metric query --disable-rbac 'node_filesystem_size_bytes{mountpoint="/"}'
```

With the `"fstype" in result.stdout` in the until condition, the grep is redundant.
With no pipes in the command, you can use the command module instead of shell. `command` is recommended instead of `shell` because it is faster.

elfiesmelfie (changes requested):

roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
jlarriba: Thank you for the suggestion, it true. I fixed it.

However, the same applies to the previous task, imho, so I have removed the grep from there too.

roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
```diff
@@ -36,4 +37,34 @@
...
+
+- name: |
+    TEST Verify node exporter filesystem metrics report a real filesystem type instead of overlay
+  ansible.builtin.command: |
+    {{ openstack_cmd }} metric query --disable-rbac 'node_filesystem_size_bytes{mountpoint="/"}'
+  register: result
+  delay: 30
+  retries: 10
+  changed_when: false
+  until: result.rc == 0 and "overlay" not in result.stdout and "fstype" in result.stdout
```
elfiesmelfie: Since these are checking the same thing, you could run the command once and check the output in two separate assert tasks.

```
- name: |
    Get the filesystem metrics
  ansible.builtin.command: |
    {{ openstack_cmd }} metric query --disable-rbac 'node_filesystem_size_bytes{mountpoint="/"}'
  register: result
  delay: 30
  retries: 10
  changed_when: false
  until: result.rc == 0

- name: |
    TEST Verify node exporter reports host filesystem mountpoints instead of container-internal ones
  ansible.builtin.assert:
      that:
          - '"mountpoint" in result.stdout'

- name: |
    TEST Verify node exporter filesystem metrics report a real filesystem type instead of overlay
  ansible.builtin.assert:
    that:
        - '"overlay" not in result.stdout'
        - '"fstype" in result.stdout'
```

You can also set a `success_msg` and `fail_msg` for `assert` [1], so  that there's a user-friendly message printed on failure, which is really useful when multiple conditions are being asserted.
[1] https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/assert_module.html

elfiesmelfie (changes requested):

roles/telemetry_verify_metrics/tasks/verify_node_exporter_metrics.yml
jlarriba: good suggestion! It makes the code more readable, implemented it in the PR

elfiesmelfie:
recheck

elfiesmelfie:
> recheck

The build history for the meta-content-provider job shows that it is working again

jlarriba:
recheck

elfiesmelfie (approved):

vyzigold (approved):

PR #338, opened 2026-03-18
[telemetry_chargeback] Add CloudKitty Scope API verification tests

jlarriba:
## Summary
- Adds verification tests for the CloudKitty Scope API POST `/v2/scope` fix (OSPRH-23754, CL 970605)
- Tests that creating a scope with `last_processed_timestamp` succeeds and returns correct data
- Tests that the created scope persists and is retrievable via GET
- Tests that omitting `last_processed_timestamp` correctly returns HTTP 400

## Test plan
- [ ] Run `ansible-lint` on the new task file
- [ ] Execute the chargeback test suite against a deployed environment with the CloudKitty fix applied
- [ ] Verify all three TEST assertions pass in the CI job output

### DIFF
--- a/roles/telemetry_chargeback/tasks/scope_api_tests.yml
+++ b/roles/telemetry_chargeback/tasks/scope_api_tests.yml
@@ -1,0 +1,98 @@
+---
+# Verification test for OSPRH-23754: CloudKitty Scope API POST fix
+# The Scope API POST /v2/scope was broken because last_processed_timestamp
+# was not accepted as input, and the active field defaulted to None.
+# CL 970605 fixed this by making last_processed_timestamp a required input
+# parameter and defaulting active to True.
+
+- name: Get OpenStack auth token
+  ansible.builtin.command:
+    cmd: "{{ openstack_cmd }} token issue -f value -c id"
+  register: os_token
+  changed_when: false
+  failed_when: os_token.rc != 0
+
+- name: Get CloudKitty rating endpoint
+  ansible.builtin.shell:
+    cmd: "{{ openstack_cmd }} endpoint list --service rating --interface internal -f value -c URL | head -1"
+  register: ck_endpoint
+  changed_when: false
+  failed_when: ck_endpoint.rc != 0 or ck_endpoint.stdout == ""
+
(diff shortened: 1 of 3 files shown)


roles/telemetry_chargeback/tasks/scope_api_tests.yml
```diff
@@ -0,0 +1,98 @@
...
+    body_format: json
+    body:
+      scope_id: "{{ ck_test_scope_id }}"
+      scope_key: "project_id"
+      fetcher: "keystone"
+      collector: "prometheus"
+      last_processed_timestamp: "{{ ck_test_scope_timestamp }}"
+    status_code: [200, 409]
+    return_content: true
+  register: scope_create_response
```
elfiesmelfie: This task is failing with the following error:

```
TASK [telemetry_chargeback : Create a scope via the Scope API POST] ************
Friday 05 June 2026  09:02:38 +0000 (0:00:00.033)       0:00:36.361 *********** 
fatal: [localhost]: FAILED! => {"changed": false, "content": "", "elapsed": 0, "msg": "Status code was -1 and not [200, 409]: Request failed: <urlopen error [Errno -2] Name or service not known>", "redirected": false, "status": -1, "url": "https://cloudkitty-internal.openstack.svc:8889/v2/scope"}
...ignoring
```

This is causing the subsequent tasks to fail.


I believe the issue is routing. The URL for the openstack cloudkitty service might not be accessible outside of the cluster by default.

There are two options that I can see:

1. We can reach the URLs from within the openstackclient pod without further config. This means converting that URI module usage into command module + `oc rsh openstack client curl ....` 
2. Alternatively, for the URI module to work, you will need to make the cloudkitty URL routable from the controller node.

elfiesmelfie (changes requested):

roles/telemetry_chargeback/tasks/scope_api_tests.yml
jlarriba: Thank you for your input, I have fixed this part in hopes that it works now.

jlarriba:
recheck

PR #335, opened 2026-02-03
[telemetry_chargeback] Get chargeback data results from loki and validate (TEMP)

ayefimov-1:
Test changes validate Synthetic data push and pull from loki
- Generates Synthetic data
- Uploads data to loki
- pull data from loki
- Validates chargeback synthetic data upload and download
- Checks to see that CK rating data is generated from synthetic data 

Closes: https://issues.redhat.com/browse/OSPRH-26066
Closes: https://issues.redhat.com/browse/OSPRH-26361
Closes: https://issues.redhat.com/browse/OSPRH-26969

Assisted-by: Claude

PR #340, opened 2026-03-24
Add the ability get the chargeback rate and diff against the calculated rate

ayefimov-1:
Depends-On:  https://github.com/infrawatch/feature-verification-tests/pull/335

Adds chargeback rate validation to the telemetry_chargeback role.  
- calculates chargeback rate for data pushed to loki and data pulled form


ayefimov-1:
recheck

PR #345, opened 2026-04-15
Add generic container-image-content-provider job

vyzigold:
Add an abstract Zuul job and playbook that builds a container image from the current change and serves it from a local registry. This provides a reusable content-provider pattern for repos like sg-core, prometheus-podman-exporter, and mysqld_exporter that need container image builds as part of the telemetry verification pipeline.

Generated-By: Claude-Code claude-opus-4-6

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -2,6 +2,7 @@
 - job:
     name: telemetry-container-image-content-provider
     parent: cifmw-base-minimal
+    nodeset: centos-stream-9-vexxhost
     description: |
       Build sg-core, prometheus-podman-exporter and mysqld-exporter
       container images from the current change and serve them from
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,5 +1,21 @@
 ---
 - job:
+    name: telemetry-container-image-content-provider
+    parent: cifmw-base-minimal
+    abstract: true
+    description: |
+      Build a container image from the current change and serve it from
+      a local registry. Child jobs must set:
+        container_build_src: path to the project source
+        container_build_image_name: image name (e.g. sg-core)
+        container_update_var_name: cifmw_update_containers variable to return
+    required-projects:
+      - name: openstack-k8s-operators/ci-framework
+        override-checkout: main
+    run:
+      - ci/playbooks/container-image-content-provider.yml
+
+- job:
     name: functional-tests-osp18
     dependencies: ["telemetry-openstack-meta-content-provider-master"]
     parent: telemetry-operator-multinode-autoscaling
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -41,7 +41,7 @@
       content_provider_dlrn_md5_hash: ''
     vars:
       patch_observabilityclient: true
-      cifmw_update_containers: false
+      cifmw_update_containers: true
       crc_enable_monitoring: true
       cifmw_extras:
         - "@{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir }}/scenarios/centos-9/multinode-ci.yml"
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -2,18 +2,29 @@
 - job:
     name: telemetry-container-image-content-provider
     parent: cifmw-base-minimal
-    abstract: true
     description: |
-      Build a container image from the current change and serve it from
-      a local registry. Child jobs must set:
-        container_build_src: path to the project source
-        container_build_image_name: image name (e.g. sg-core)
-        container_update_var_name: cifmw_update_containers variable to return
+      Build sg-core, prometheus-podman-exporter and mysqld-exporter
+      container images from the current change and serve them from
+      a local registry for dependent jobs.
     required-projects:
       - name: openstack-k8s-operators/ci-framework
         override-checkout: main
+      - name: openstack-k8s-operators/sg-core
+      - name: openstack-k8s-operators/prometheus-podman-exporter
+      - name: openstack-k8s-operators/mysqld_exporter
     run:
       - ci/playbooks/container-image-content-provider.yml
+    vars:
+      container_images:
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/sg-core"
+          name: sg-core
+          update_var: cifmw_update_containers_ceilometersgcoreImage
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/prometheus-podman-exporter"
+          name: prometheus-podman-exporter
+          update_var: cifmw_update_containers_edpmpodmanexporterImage
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/mysqld_exporter"
+          name: mysqld-exporter
+          update_var: cifmw_update_containers_ceilometermysqldexporterImage
 
 - job:
     name: functional-tests-osp18
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -22,8 +22,41 @@
           - crc
 
 - job:
+    name: telemetry-container-image-content-provider
+    parent: cifmw-base-minimal
+    description: |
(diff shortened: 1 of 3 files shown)


ci/playbooks/container-image-content-provider.yml
```diff
@@ -0,0 +1,51 @@
...
+      ansible.builtin.set_fact:
+        _container_image_url: "{{ cifmw_rp_registry_ip }}:5001/{{ container_build_image_name }}:{{ _container_image_tag }}"
+
+    - name: Build container image
+      ansible.builtin.command:
+        cmd: >-
+          make docker-build
+          IMG={{ _container_image_url }}
+      args:
+        chdir: "{{ container_build_src }}"
```
elfiesmelfie: Why `args: chdir` instead of chdir, which is a supported arg of the command module.



```suggestion
        chdir: "{{ container_build_src }}"
```

Same for multiple instances in this file

ci/playbooks/container-image-content-provider.yml
```diff
@@ -0,0 +1,51 @@
...
+      register: _container_sha
+
+    - name: Set image tag
+      ansible.builtin.set_fact:
+        _container_image_tag: "{{ _container_sha.stdout | trim }}"
+
+    - name: Set full image reference
+      ansible.builtin.set_fact:
+        _container_image_url: "{{ cifmw_rp_registry_ip }}:5001/{{ container_build_image_name }}:{{ _container_image_tag }}"
```
elfiesmelfie: Does this only support single container image?

Is there a usecase where we'd need to build multiple containers? e.g. with multiple depends-on?

elfiesmelfie (changes requested):

elfiesmelfie:
- I would like to see a concrete child job running before this merges

ci/playbooks/container-image-content-provider.yml
vyzigold: Yes, it supported only a single image, which was my original goal. But I think it's a good idea to just build all 3 of the images. Although I don't think we'll need to build more than 1 anytime soon, it'll be a bit more future-proof. 

The 3rd commit rewrote the job to build all 3 images.

vyzigold:
> * I would like to see a concrete child job running before this merges

This is a bit of an issue, which is why I wanted to merge this PR and update the job as we start to use it. I'm pretty sure you need the job to exist outside of a PR branch, before you can define child jobs. But for now I solved it with just temporarily duplicating the whole job into my sg-core PR (https://github.com/openstack-k8s-operators/sg-core/pull/33) . I'll try to keep this PR and the sg-core one in sync and once we see the content provider working there, we'll have a pretty good proof that this PR works.

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
recheck nodefail

vyzigold:
request nodefailure

vyzigold:
recheck nodefailure

vyzigold:
recheck

vyzigold:
recheck

vyzigold:
All the changes this previously depended on are merged now. I removed the Depends-Ons and rebased on top of master, let's get a clean run in Zuul, afterwards this should be ready for reviews.

vyzigold:
recheck failed netconfig deploy, which caused a retried not idempotent enable audit logging to fail as well

ci/playbooks/container-image-build.yml
```diff
@@ -0,0 +1,28 @@
...
+    cmd: >-
+      make docker-build
+      IMG={{ _container_image_url }}
+    chdir: "{{ container_image.src }}"
+
+- name: "Push {{ container_image.name }}"
+  ansible.builtin.command:
+    cmd: >-
+      make docker-push
+      IMG={{ _container_image_url }}
```
danpawlik: the `IMG={{ _container_image_url }}` should not be before make docker-push?

ci/playbooks/container-image-content-provider.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Build and serve container images from review
+  hosts: "{{ cifmw_target_host | default('controller') }}"
+  gather_facts: true
+  tasks:
+    - name: Discover registry IP
+      ansible.builtin.set_fact:
+        cifmw_rp_registry_ip: "{{ ansible_host }}"
```
danpawlik: wondering if `{{ ansible_default_ipv4.address }}` would not be better here

.zuul.yaml
```diff
@@ -21,9 +21,42 @@
...
+      container_images:
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/sg-core"
+          name: sg-core
+          update_var: cifmw_update_containers_ceilometersgcoreImage
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/prometheus-podman-exporter"
+          name: prometheus-podman-exporter
+          update_var: cifmw_update_containers_edpmpodmanexporterImage
+        - src: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/mysqld_exporter"
+          name: mysqld-exporter
+          update_var: cifmw_update_containers_ceilometermysqldexporterImage
```
danpawlik, line 52: the `ceilometer` here is correct?

vyzigold: Unfortunately yes. This is related to how the controlplane CRD (and the telemetry controller) is structured. Originally we started with just one exporter, which was ceilometer. So we have a telemetry.ceilometer section in the CRD. Later we started adding exporters (for the controlplane that'd be kube-state-metrics and mysqld_exporter). We decided to keep them together in the CRD and we didn't want to add new "top-level" sections in the CRD. Which resulted in various config fields related to kube-state-metrics and mysqld_exporter being under "ceilometer". This hierarchy is then inherited in the openstackversions CRD, which controls which images are used for which service, in there we have the `ceilometerMysqldExporterImage` field, which is where this variable comes from.

Unfortunately we're stuck with this strange naming. We may be able to change it for RHOSO 19, but I personally don't think that's gonna happen. In cases like this, you can imagine the word "ceilometer" being "exporter"

ci/playbooks/container-image-content-provider.yml
vyzigold: Why? I'll try to research what the differences could be. I'd assume the vars would be pretty similar except for ansible_host maybe working in IPv6 as well? :thinking:

.zuul.yaml
danpawlik: Make sense. I was thinking that it is just copy-paste error

ci/playbooks/container-image-content-provider.yml
vyzigold: I'll use https://github.com/openstack-k8s-operators/ci-framework/blob/81ed403edc15e23cc5e4e268e5dffe3891f0b9d5/ci/playbooks/content_provider/pre.yml#L22 as discussed on slack

ci/playbooks/container-image-build.yml
vyzigold: The order actually doesn't matter, both work. But you have a point, this goes against convention and may be slightly confusing when reading it. I'll reverse the order since I'm doing other changes.

vyzigold: done

ci/playbooks/container-image-content-provider.yml
vyzigold: done

danpawlik (dismissed):
So far, LGTM. Let's wait for the CI

vyzigold:
recheck

ci/playbooks/container-image-build.yml
vyzigold: Interesting. The ansible.builtin.command treats the first part of the first line as the executable. So it's trying to execute the `IMG=...` and that's failing. I'll return the order back, since that worked.

danpawlik (approved):

elfiesmelfie (approved):

PR #362, opened 2026-04-21
[telemetry_chargeback] Task files needed to run FVT test with multiple test scenarios

ayefimov-1:
Task files need to running multiple test scenarios for telemetry_chargeback role
- Allows for multiple test scenario files 
- Synthetic data is generated for each scenario in artifacts dir
- Synthetic data summary is generated for each test scenario in artifacts dir

Written by @ayefimov-1 
Assisted-by: Claude


myadla:
recheck

roles/telemetry_chargeback/tasks/run_test_scenarios.yml
elfiesmelfie, line 1: Why is this task separated into its own file?

if it's on it's own for now, please add a placeholder comment or debug task that indicates there will be more test-steps later

ayefimov-1: Its on its own because of the looping needed for multiple test scenarios.  See tasks/main.yml L#44 

This tasks file will have more tasks by the time this role is done.

ayefimov-1:
recheck

vyzigold:
I asked Claude to review this. Here is what it found and I agree with:

- vars/main.yml conflicts with defaults/main.yml. vars/main.yml still defines logs_dir_zuul and artifacts_dir_zuul`
- --user-id is never passed to the script. In gen_synth_loki_data.yml, the command passes -p for project ID but never passes -u for user ID, even though cloudkitty_user_id is computed in main.yml
- In gen_synth_loki_data.yml, --debug "{{ cloudkitty_debug_dir }}" is passed to gen_db_summary.py (a different script not in this PR). But the existing gen_synth_loki_data.py has --debug as a boolean flag. Would be good to unify the behavior of the flag between scripts or use a different name.

ayefimov-1:
> I asked Claude to review this. Here is what it found and I agree with:
> 
> * vars/main.yml conflicts with defaults/main.yml. vars/main.yml still defines logs_dir_zuul and artifacts_dir_zuul`

Removed from vars/main.yml

> * --user-id is never passed to the script. In gen_synth_loki_data.yml, the command passes -p for project ID but never passes -u for user ID, even though cloudkitty_user_id is computed in main.yml

Both removed (these were development aids for me)
               
> * In gen_synth_loki_data.yml, --debug "{{ cloudkitty_debug_dir }}" is passed to gen_db_summary.py (a different script not in this PR). But the existing gen_synth_loki_data.py has --debug as a boolean flag. Would be good to unify the behavior of the flag between scripts or use a different name.

This was done intentionally as each have very different goals and would prefer keeping these as is: 

Debug option for gen_db_summary.py is intended for runtime debugging and generates too much information to send to stdout.  So, the debugging info is sent to {{ cloudkitty_debug_dir }} and generates a separate file for each test scenario. 

Debug for gen_synth_loki_data.py is non-runtime/runtime debug info to stdout.

ayefimov-1:
recheck

vyzigold:
> > * In gen_synth_loki_data.yml, --debug "{{ cloudkitty_debug_dir }}" is passed to gen_db_summary.py (a different script not in this PR). But the existing gen_synth_loki_data.py has --debug as a boolean flag. Would be good to unify the behavior of the flag between scripts or use a different name.
> 
> This was done intentionally as each have very different goals and would prefer keeping these as is:
> 
> Debug option for gen_db_summary.py is intended for runtime debugging and generates too much information to send to stdout. So, the debugging info is sent to {{ cloudkitty_debug_dir }} and generates a separate file for each test scenario.
> 
> Debug for gen_synth_loki_data.py is non-runtime/runtime debug info to stdout.

~~The thing is, we have a 2 similarly named scripts in the same folder having the same named arg, but each behaves completely different. What about having --debug-dir in gen_db_summary.py~~

EDIT: Or actually that's an already merged code, so forget that, let's not pollute this PR

roles/telemetry_chargeback/tasks/main.yml
```diff
@@ -1,6 +1,37 @@
...
   ansible.builtin.include_tasks: "chargeback_tests.yml"
 
-- name: "Generate Synthetic Data"
-  ansible.builtin.include_tasks: "gen_synth_loki_data.yml"
+- name: "Setup Loki Environment"
+  ansible.builtin.include_tasks: "setup_loki_env.yml"
+
+- name: "CloudKitty debug ON/OFF"
+  ansible.builtin.set_fact:
+    cloudkitty_debug_dir: "{{ (cloudkitty_debug | bool) | ternary(artifacts_dir_zuul + '/debug_ck_db', '') }}"
```
elfiesmelfie: Move this into defaults/main.yml
It can be always set. Since I expect that it's only used when cloudkitty_debug is true, then it's reasonable to leave it in defaults.

If there are any tasks that are conditionally executing based on the value of this cloudkitty_debug_dir variable, then they should be modified to run conditionally based on whether cloudkitty_debug is set.

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

elfiesmelfie:
recheck
Merged: #Added depends-on for https://github.com/infrawatch/feature-verification-tests/pull/366

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

TristanCacqueray:
recheck

TristanCacqueray:
recheck

ayefimov-1:
recheck

elfiesmelfie:
recheck

elfiesmelfie (dismissed):
Looks okay, but we need to wait for CI to pass.

For future PRs, here are afew things to consider to make smaller changes: 
 
- The variable renames and moving from vars/ to defaults/ could have been a separate, easier to review change, since it is a cleanup separate from the main work.
- You can make changes to the Python scripts separately to the Ansible changes. As long as the new changes are opt-in, then the Ansible can be changed separately to the Python scripts. 
- Alternatively, if you want to split out the python, you can add each command line arg or new functionality added one at a time, and incorporate all the changes from the Python update, to the Ansible and any CI changes that are needed in a single PR.

ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,30 @@
 ---
+# OpenStack CLI command
 openstack_cmd: "openstack"
+
+# Debug mode set
+cloudkitty_debug: false
+cloudkitty_debug_dir: "{{ (cloudkitty_debug | bool) | ternary(artifacts_dir_zuul + '/debug_ck_db', '') }}"
+
+# Directory paths
+logs_dir_zuul: "{{ ansible_env.HOME }}/ci-framework-data/logs"
```
danpawlik: nit: normally after `gather_facts` the `ansible_env.HOME` is set to `{{ ansible_user_dir }}`
From the other side, the playbook if is executed by ci-framework, it should use group_vars - https://github.com/openstack-k8s-operators/ci-framework/blob/main/group_vars/all.yml

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,30 @@
...
+
+# Debug mode set
+cloudkitty_debug: false
+cloudkitty_debug_dir: "{{ (cloudkitty_debug | bool) | ternary(artifacts_dir_zuul + '/debug_ck_db', '') }}"
+
+# Directory paths
+logs_dir_zuul: "{{ cifmw_basedir }}/logs"
+artifacts_dir_zuul: "{{ cifmw_basedir }}/artifacts"
+cert_dir: "{{ ansible_user_dir }}/ck-certs"
+local_cert_dir: "{{ cifmw_basedir }}/flush_certs"
```
vyzigold: Any reason for having different paths for the certs?

vyzigold: why is one inside ansible_user_dir antd the other inside cifmw_basedir?

roles/telemetry_chargeback/files/gen_synth_loki_data.py
```diff
@@ -175,37 +214,49 @@ def generate_loki_data(
...
+            raise ValueError("Each log type must have a 'type' field")
 
         # Validate required fields
-        missing = [f for f in required_fields if f not in log_type_config]
+        # metadata is optional for generation; name is not a log-type field
+        required_for_item = [
+            f for f in required_fields
+            if f not in ("name", "metadata")
+        ]
+        missing = [f for f in required_for_item if f not in log_type_config]
```
vyzigold:
```suggestion
        required_for_item = set(required_fields) - {"name", "metadata"}
        missing = required_for_item - set(log_type_config)
```
I don't know, maybe it's just me not being used to list comprehensions, but it always take a minute for me to understand what'll be the result especially with conditions involved. I see both are basically creating a list from another list without some elements. We could think about at as subtracting 2 sets, which is IMO cleaner. But your call, I won't block merging based on this. The only difference here is, that sets aren't ordered, which I don't think matters here.

roles/telemetry_chargeback/README.md
```diff
@@ -33,31 +33,61 @@ The role uses the following variables to control the testing environment and exe
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 | `openstack_cmd` | `openstack` | The command used to execute OpenStack CLI calls. This can be customized if the binary is not in the standard PATH. |
+| `cloudkitty_debug` | `false` | Enable debug mode for CloudKitty database dumps. |
+| `logs_dir_zuul` | `{{ ansible_env.HOME }}/ci-framework-data/logs` | Directory for log files. |
+| `artifacts_dir_zuul` | `{{ ansible_env.HOME }}/ci-framework-data/artifacts` | Directory for generated artifacts and test output. |
+| `cert_dir` | `{{ ansible_user_dir }}/ck-certs` | Directory for CloudKitty client certificates. |
+| `local_cert_dir` | `{{ ansible_env.HOME }}/ci-framework-data/flush_certs` | Local directory for certificate extraction. |
+| `cloudkitty_namespace` | `openstack` | Kubernetes namespace where CloudKitty is deployed. |
```
vyzigold: cloudkitty_debug_dir missing.

roles/telemetry_chargeback/files/gen_synth_loki_data.py
vyzigold: We could also be more explicit and make it more clear what's happening with `missing = required_for_item - set(log_type_config.keys())`

roles/telemetry_chargeback/defaults/main.yml
danpawlik: ah, you mean `cert_dir` and `local_cert_dir` - good catch.

elfiesmelfie: I **think** one of them is in the openstackclient pod IINM. This is something that @myadla might be able to answer!

roles/telemetry_chargeback/README.md
```diff
@@ -33,31 +33,61 @@ The role uses the following variables to control the testing environment and exe
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 | `openstack_cmd` | `openstack` | The command used to execute OpenStack CLI calls. This can be customized if the binary is not in the standard PATH. |
+| `cloudkitty_debug` | `false` | Enable debug mode for CloudKitty database dumps. |
+| `logs_dir_zuul` | `{{ ansible_env.HOME }}/ci-framework-data/logs` | Directory for log files. |
+| `artifacts_dir_zuul` | `{{ ansible_env.HOME }}/ci-framework-data/artifacts` | Directory for generated artifacts and test output. |
+| `cert_dir` | `{{ ansible_user_dir }}/ck-certs` | Directory for CloudKitty client certificates. |
+| `local_cert_dir` | `{{ ansible_env.HOME }}/ci-framework-data/flush_certs` | Local directory for certificate extraction. |
```
elfiesmelfie: If this is just for extraction, then a temp dir would be better, and wouldn't need to be included here.

If the dir is also for saving the certs for later, then add this usage also.

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -0,0 +1,56 @@
...
+  register: loki_route
+  changed_when: false
+
+- name: "Set Loki URLs"
+  ansible.builtin.set_fact:
+    # Base URL
+    loki_base_url: "https://{{ loki_route.stdout }}"
+
+    # Internal Flush URL (Service DNS: https://<service>.<namespace>.svc:3100/flush)
+    ingester_flush_url: "https://cloudkitty-lokistack-ingester-http.{{ cloudkitty_namespace }}.svc:3100/flush"
```
elfiesmelfie, line 18: Where is the ingester_flush_url used?  Is it just for the debug message?

roles/telemetry_chargeback/defaults/main.yml
myadla: Yes @elfiesmelfie. "cert_dir" is the one where gateway certs are written to and "local_cert_dir" is where openstackclient pod/ingester certs are written to. 

@vyzigold Yes, both can be created under same "ansible_user_dir" directory.

ayefimov-1:
recheck

myadla (dismissed):
looks good

ayefimov-1:
recheck

elfiesmelfie:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -1,2 +1,30 @@
...
+# Debug mode set
+cloudkitty_debug: false
+cloudkitty_debug_dir: "{{ (cloudkitty_debug | bool) | ternary(artifacts_dir_zuul + '/debug_ck_db', '') }}"
+
+# Directory paths
+logs_dir_zuul: "{{ cifmw_basedir }}/logs"
+artifacts_dir_zuul: "{{ cifmw_basedir }}/artifacts"
+cert_dir: "{{ cifmw_basedir }}/ck-certs"
+local_cert_dir: "{{ cifmw_basedir }}/flush_certs"
+remote_cert_dir: "osp-certs"
```
elfiesmelfie, line 14: cert_dir, local_cert_dir and remote_cert_dir? What are they all for?
Can you add a short description of each?

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -0,0 +1,53 @@
...
+    # Base URL
+    loki_base_url: "https://{{ loki_route.stdout }}"
+
+- name: "Set Derived Loki URLs"
+  ansible.builtin.set_fact:
+    loki_push_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/push"
+    loki_query_url: "{{ loki_base_url }}/api/logs/v1/cloudkitty/loki/api/v1/query_range"
+
+- name: "Debug URLs"
+  ansible.builtin.debug:
```
elfiesmelfie, line 26: Is this the only place that loki_query_url is used?

You could use a var for the task instead, to stop facts leaking, and make it easier to grok scope of the variable. Or even omit it from the debug, if it is not relevant

roles/telemetry_chargeback/README.md
```diff
@@ -1,81 +1,333 @@
...
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-| `openstack_cmd` | `openstack` | The command used to execute OpenStack CLI calls. This can be customized if the binary is not in the standard PATH. |
+| `openstack_cmd` | `"openstack"` | OpenStack CLI command (customize if not in PATH) |
+| `cloudkitty_debug` | `false` | Enable debug mode for CloudKitty operations |
+| `cloudkitty_debug_dir` | `"{{ (cloudkitty_debug \| bool) \| ternary(artifacts_dir_zuul + '/debug_ck_db', '') }}"` | Directory for debug output (auto-set based on debug flag) |
+| `logs_dir_zuul` | `"{{ cifmw_basedir }}/logs"` | Directory for log files |
+| `artifacts_dir_zuul` | `"{{ cifmw_basedir }}/artifacts"` | Directory for generated artifacts and test output |
+| `cert_dir` | `"{{ ansible_user_dir }}/ck-certs"` | Directory for CloudKitty client certificates |
+| `local_cert_dir` | `"{{ cifmw_basedir }}/flush_certs"` | Local directory for flush certificates (cleaned up after run) |
```
elfiesmelfie: If this is cleaned up after the run, it doesn't need to be something that the user can configure. 

Use tempfile to create a temp file for the flush certs

roles/telemetry_chargeback/README.md
```diff
@@ -1,81 +1,333 @@
...
 
-| Variable | Default Value | Description |
-|----------|---------------|-------------|
-| `logs_dir_zuul` | `/home/zuul/ci-framework-data/logs` | Remote directory for log files. |
-| `artifacts_dir_zuul` | `/home/zuul/ci-framework-data/artifacts` | Directory for generated artifacts. |
-| `cloudkitty_synth_script` | `{{ role_path }}/files/gen_synth_loki_data.py` | Path to the synthetic data generation script. |
-| `cloudkitty_data_template` | `{{ role_path }}/templates/loki_data_templ.j2` | Path to the Jinja2 template for Loki data format. |
-| `ck_data_config` | `{{ role_path }}/files/test_static.yml` | Path to the scenario configuration file. |
-| `ck_output_file_local` | `{{ artifacts_dir_zuul }}/loki_synth_data.json` | Local path for generated synthetic data. |
-| `ck_output_file_remote` | `{{ logs_dir_zuul }}/gen_loki_synth_data.log` | Remote destination for synthetic data. |
```
elfiesmelfie, line 49: I thought these were removed in a separate change? I'm surprised to see this variable list on master.

This is not relevant to the review, but I'm glad they're getting cleaned up

roles/telemetry_chargeback/README.md
```diff
@@ -1,81 +1,333 @@
...
+   - Calculates expected chargeback metrics (`gen_db_summary.py`)
+   - Loads metrics for validation
+
+5. **Cleanup** (`cleanup_ck.yml`)
+   - Removes temporary certificate directories
+   - Always runs (even on failure) via block/rescue/always structure
+
+### Loop Variable
+
+The role uses `{{ scenario_name }}` as the loop variable when processing multiple test scenarios, making it easy to track which scenario is currently executing.
```
elfiesmelfie: This is not really relevant and ideally should be removed. But this can be done as a follow-up

elfiesmelfie (changes requested):
If I was doing this, I'd consider splitting out the documentation changes, since they cover the python scripts. Ideally, the documentation updates would be added when the python scripts were being added, I'm not going to block on that particular point, but please keep it in mind in future. 

There are a few questions that I'd like to see addressed, but once they're resolved, I'll approve

ayefimov-1:
recheck

ayefimov-1:
recheck

elfiesmelfie (approved):

ayefimov-1:
recheck

ayefimov-1:
recheck

danpawlik (approved):

PR #364, opened 2026-04-23
[telemetry_chargeback] Add ability to retrieve loki data

ayefimov-1:
Add the ability to retrieve loki data from database
- Pulls data from loki
- Does basic validation of retrieved  loki data

Authored-by: @myadla
Co-authored-by: @ayefimov-1
AI Assisted by: Claude


ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

myadla (approved):
lgtm

danpawlik (approved):

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
```diff
@@ -0,0 +1,71 @@
...
+        validate_certs: false
+        return_content: true
+        body_format: json
+      register: loki_response
+      # Wait condition
+      until:
+        - loki_response.status == 200
+        - loki_response.json.status == 'success'
+        - loki_response.json.data.result | length > 0
+        - (loki_response.json.data.result | map(attribute='values') | map('length') | sum) >= (synth_data_rates.data_log.log_count | int)
```
vyzigold: I'm pretty sure I've pointed out in previous PRs, that this condition may have issues when multiple test scenarios are executed after each other. I wonder if it was testing with multiple test scenarios?

What happens if:
```
1. test_static executes
2.     as part of that let's say 10 logs get pushed into loki
3.     The ansible waits on this task until all 10 logs are inside loki. This is correct and all works
4. Now test_bin_basic starts executing
5.     as part of that another 10 logs get pushed into loki on top of the previous 10 that are already there from the previous scenario
6.     This task immediately succeeds and the ansible moves on, because we already have 10 logs in loki from previous scenario, which is >= log_count of the current scenario. It doesn't care, that some of the logs from the current scenario didn't get returned yet.
```

vyzigold: I know I've seen that issue before and maybe you've fixed it somehow, but I may have missed the fix

danpawlik: does it mean that condition in  L38 should also be changed?

ayefimov-1: This has been run with multiple scenarios and does not seem to be an issue ...at least for now. The job tests pushes loki data to the DB, then it pulls data from the DB and finally it compares the pushed and pulled data. If it is different then the job fails.

Having said all of that, to your point, I don't understand why it does not fail. 

What I would like is more a query with a beginning and ending timestep:

url: "{{ loki_query_url }}?query={{ logql_query | urlencode }}&start={{ synth_data_rates.time.begin_step.nanosec }}&end={{synth_data_rates.time.end_step.nanosec }}&limit={{ limit }}"

roles/telemetry_chargeback/README.md
```diff
@@ -40,53 +43,151 @@ These variables can be overridden when importing the role or set at the play lev
...
+| `remote_cert_dir` | `"osp-certs"` | Remote directory inside OpenStack pod for certificates |
+| `cert_secret_name` | `"cert-cloudkitty-client-internal"` | OpenShift secret name for client certificates |
+| `client_secret` | `"secret/cloudkitty-lokistack-gateway-client-http"` | Secret for flush client certificates |
+| `ca_configmap` | `"cm/cloudkitty-lokistack-ca-bundle"` | ConfigMap for CA bundle |
+| `logql_query` | `"{service=\"cloudkitty\"}"` | LogQL query for Loki (overridable via `loki_query`) |
+| `cloudkitty_namespace` | `"openstack"` | Kubernetes namespace where CloudKitty is deployed |
+| `openstackpod` | `"openstackclient"` | OpenStack client pod name for exec/cp operations |
+| `lookback` | `6` | Days to look back for Loki query time range |
+| `limit` | `50` | Limit for Loki query results |
+| `cloudkitty_test_scenarios` | `[]` | List of test scenario files to run (empty = auto-discover) |
```
elfiesmelfie: These changes to this part of the README revert the changes in the parent PR, i.e. reverting to the state two PRs ago. 

This needs to be addressed

elfiesmelfie (changes requested):

roles/telemetry_chargeback/tasks/retrieve_loki_data.yml
vyzigold: that query makes sense. Especially if the data from the scenarios don't overlap timewise. But if you've done tests and as you say, the data retrieved from Loki match what has been pushed, then we're probably OK.

ayefimov-1: @vyzigold  Or another another approach ...  

Each test scenario gets its own isolated Loki label, preventing cross-contamination when multiple scenarios run sequentially. And i can differentiate multiple runs of the same scenario with multiple runs by start label.

vyzigold: That sounds like a good solution to me.

PR #365, opened 2026-04-27
DNM testing CI

vyzigold:
Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/877


vyzigold:
Issue was the dependend PR was disabling Swift, which is being used by glance as a storage backend. So image creation would fail. We have a different fix, so I'm closing this.

PR #367, opened 2026-04-27
[telemetry_chargeback] Make variable names and values consistent within the role and with prometheus and loki (TEMP)

ayefimov-1:
Made changes to the python scripts for consistency and accuracty:
- Changed user, project and resource variable consisent with loki/prometheus
- Make python script params more consistent
- fixed python files to handle mutate: NUMBOOL correctly when calculating rating
- add scenario name to the db push so that will be target db data to pull

Closes:
[OSPRH-29232](https://redhat.atlassian.net/browse/OSPRH-29232)

Authored-by @ayefimov-1 
Assisted-by: Claude


ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/files/gen_db_summary.py
elfiesmelfie, line 1: Correct me if I'm wrong, but the last change to this file was on master, right?
If this is the case, then the update to this file can be proposed directly against master

roles/telemetry_chargeback/files/test_dyn_basic.yml
elfiesmelfie, line 1: The changes to this file also appear to be the first changes since master. Do the changes here require all the other PRs in order to work?
If no, then this could be proposed against master.

elfiesmelfie (commented):
This looks like minor changes that are independent from the other changes to the telemetry_chargeback role.

I **think ** that this could be proposed against master and merged quickly, since it is a straightforward-seeming change.

Certainly, the test*.yml files look like they're safe to propose separately.
The python file changes should ideally be proposed separately from the test*yml ones, but the changes could be merged together.

Please re-target this PR against master, if you agree with my assessment, and it makes sense. Then ping me and I'll re-review and approve!

roles/telemetry_chargeback/files/gen_db_summary.py
ayefimov-1: Yes, you are correct.  Change to into branch

roles/telemetry_chargeback/files/test_dyn_basic.yml
ayefimov-1: Correct

PR #369, opened 2026-04-29
Remove noop parent

TristanCacqueray:
This change fixes the following zuul config-error:

  "noop" is final and can not act as a parent

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -138,7 +138,6 @@
 
 - job:
     name: feature-verification-tests-noop
-    parent: noop
     description: |
       A job that always passes. Runs when there's a change to jobs that don't
       need full zuul to run but still need to report a pass.
@@ -241,4 +240,3 @@
               # case the job definition changes.
               - ^roles/telemetry_verify_metrics/tasks/verify_ceilometer_volume_pool_metrics.yml$
               - ^.zuul.yaml$
-


TristanCacqueray:
recheck

TristanCacqueray:
recheck

.zuul.yaml
```diff
@@ -138,7 +138,6 @@
 
 - job:
     name: feature-verification-tests-noop
-    parent: noop
```
elfiesmelfie, line 141: I don't understand ho it's been working as long as it has if noop is final. 
Was this change recent?

TristanCacqueray: Yes, this was changed yesterday in Zuul.

TristanCacqueray:
recheck

.zuul.yaml
vyzigold: @elfiesmelfie https://review.opendev.org/c/zuul/zuul/+/986659

vyzigold:
recheck

elfiesmelfie:
recheck

elfiesmelfie:
recheck

There are no logs, need to get a new build

elfiesmelfie:
recheck

vyzigold:
recheck POST_FAILURE with no logs, but logs are available for the other builds.

TristanCacqueray:
recheck

vyzigold:
@TristanCacqueray We discovered we had another issue with the periodic ceph job as well. And we needed to fix that and the noop parent at the same time. So I included your commit in https://github.com/infrawatch/feature-verification-tests/pull/379 . That PR merged, so this one isn't needed anymore.

Thank you a lot for proactively looking at this and proposing a fix!

PR #373, opened 2026-05-01
Adding loki flush tasks file.

ayefimov-1:
Flushes Loki Ingester Memory to Storage
- Adding tasks file to flush loki memory to storage
- Allows data to be pulled from loki with feature in future PR

Authored-by: @myadla
Co-authored-by: @ayefimov-1
AI Assisted by: Claude


vyzigold (approved):

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,52 @@
+---
+# Flush Loki Ingester Memory to Storage
+
+- name: "Flush execution inside OpenStack CLI"
+  block:
+    # create dir
+    - name: "Create directory inside OpenStack CLI"
```
elfiesmelfie:
```suggestion
    - name: "Create directory inside OpenStack Client pod"
```

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,52 @@
...
+        path: "{{ local_cert_dir }}"
+        state: directory
+        mode: '0755'
+
+    # copy all certs
+    - name: "Copy certificates to OpenStack CLI"
+      ansible.builtin.command:
+        cmd: "oc cp {{ local_cert_dir }}/. {{ cloudkitty_namespace }}/{{ openstackpod }}:{{ remote_cert_dir }}/"
+      changed_when: true
+
```
elfiesmelfie: Am I missing something? It seems that we create the empty local_cert dir and try to copy the contents to the openstackclient pod. But if the local_cert_dir was just created, then there's nothing in there to copy into the pod.

If the directory has to be created in the "create directory to extract certificates" task, then there's a problem.

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,52 @@
...
+        fail_msg: "Flush failed"
+        success_msg: "Ingester Memory Flushed successfully"
+
+  rescue:
+    - name: "Debug failure output"
+      ansible.builtin.debug:
+        msg:
+          - "Failure"
+          - "Stdout: {{ flush_response.stdout | default('') }}"
+          - "Stderr: {{ flush_response.stderr | default('') }}"
```
elfiesmelfie: This limits the output, maybe just show the whole task output i.e. `{{ flush_response }}`, since it's conceivable that the issue is in the  input  to the curl command

roles/telemetry_chargeback/tasks/ingest_loki_data.yml
elfiesmelfie, line 28: Combine these into a single task.

However, I am concerned that this will break the test execution if it's allowed to explicitly fail here. If these tests fail, then the playbook running the test roles will also fail, and any test that are configured to run after this one will not be run.

If all the TEST tasks to this point are successful, then the test report will show success since there are no TEST tasks that failed. -- this would be very misleading

roles/telemetry_chargeback/README.md
```diff
@@ -18,168 +27,66 @@ Requirements
...
-|----------|-------------|
-| `-j, --json PATH` | Input JSON file (Loki format or synthetic data) |
-
-**Optional Arguments:**
-| Argument | Default | Description |
-|----------|---------|-------------|
-| `-o, --output PATH` | `<input_stem>_total.yml` | Output YAML file path |
-| `--debug` | `False` | Enable debug mode (writes `<stem>_diff.txt` file) |
-| `--debug_dir DIR` | Output directory | Directory for debug files (defaults to output file's directory) |
-
```
elfiesmelfie, line 151: Why is the documentation for the python scripts removed in this PR? It was just added in the previous one.
If you don't want it to be added, then it should be removed completely from the commit that introduced it.

roles/telemetry_chargeback/tasks/flush_loki_data.yml
vyzigold: hmm. you're probably right. I somehow read it as "create a remote dir in the openstackclient pod and then copy the certs there". But that's not what's happening.

vyzigold: oh. I see. The remote dir is created by the first task in this file. The local_cent_dir is already present, it'll get created by this task https://github.com/infrawatch/feature-verification-tests/pull/373/changes#diff-159ee438942623db2b06b8440ca6005f864c6bf3f60b33c6b89e6ee24967f80aR46 . So I think the "Create directory to extract certificates" task probably doesn't do anything and it may not be needed.

ayefimov-1: Fixed with next push:
- Added cert_dir and local_cert_dir to be creates in the setup_loki_evn.yml 
- Removed it from the flush_loki_data.yml

roles/telemetry_chargeback/tasks/flush_loki_data.yml
ayefimov-1: Fixed in next push.

roles/telemetry_chargeback/README.md
ayefimov-1: I think this happened as a consequesce of me adjusting to the revert and new PR for  revert-372-alexy_ck_job_2 branch into alexy_db_download_tasks_files.

roles/telemetry_chargeback/tasks/flush_loki_data.yml
ayefimov-1: fixed in next push.

roles/telemetry_chargeback/tasks/ingest_loki_data.yml
ayefimov-1: What is the benefit of running the rest of the job if this part fails?  Downstream of this task, all the "test" are no longer useful. 

Fixed in next push

ayefimov-1:
recheck

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,42 @@
...
+          --cacert {{ remote_cert_dir }}/service-ca.crt
+      register: flush_response
+      changed_when: true
+      any_errors_fatal: true
+
+    # Status
+    - name: "Verify flush status"
+      ansible.builtin.assert:
+        that:
+          - "'204' in flush_response.stderr or '200' in flush_response.stderr"
```
danpawlik: is it really `stderr` or `stdout`?

danpawlik (dismissed):

roles/telemetry_chargeback/tasks/flush_loki_data.yml
ayefimov-1: Fixed in next push:

- Replaced -v with -s -w '%{http_code}' to capture status code in stdout
- Changed match to flush_response.stdout in ['200', '204']                                                                    
- Enhanced fail_msg to display actual status code received

roles/telemetry_chargeback/tasks/flush_loki_data.yml
```diff
@@ -0,0 +1,42 @@
...
+      register: flush_response
+      changed_when: true
+      any_errors_fatal: true
+
+    # Status
+    - name: "Verify flush status"
+      ansible.builtin.assert:
+        that:
+          - "flush_response.stdout in ['200', '204']"
+        fail_msg: "Flush failed with status: {{ flush_response.stdout }}"
```
danpawlik, line 36: in fail, here probably should be stderr not stdout. Can be fixed later.

danpawlik (approved):

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -30,12 +30,15 @@
...
   ansible.builtin.file:
-    path: "{{ cert_dir }}"
+    path: "{{ item }}"
     state: directory
-    mode: '0755'
+    mode: '0700'  # Restricted for security
+  loop:
+    - "{{ cert_dir }}"
+    - "{{ local_cert_dir }}"
+  when: item is defined
```
danpawlik, line 41: that condition is not needed :)

lnatapov (approved):

PR #374, opened 2026-05-04
Use base zuul noop job

vyzigold:
Use a noop job from zuul instead of a custom variant. Apparently the noop job was made final recently, so we can't inherit from it anymore.

See zuul/model.py in https://review.opendev.org/c/zuul/zuul/+/986659


vyzigold:
This is already being worked on in https://github.com/infrawatch/feature-verification-tests/pull/369

PR #377, opened 2026-05-06
Adding database upload tasks file ingest_loki_data.yml

ayefimov-1:
This file pushes synthetic data generated during a FVT run to loki

- Add the ability to ingest and retrieve loki data from database
- Uploads data to loki

Authored-by: @myadla
Co-authored-by: @ayefimov-1
AI Assisted by: Claude


danpawlik (approved):
Normally it is fine, but might be helpful to add into the commit message why it has been reverted

ayefimov-1:
> Normally it is fine, but might be helpful to add into the commit message why it has been reverted

I inadvertently merge the change.  So, just a "miss-click"

elfiesmelfie (commented):
There is no difference between this branch and the target branch

ayefimov-1:
> There is no difference between this branch and the target branch

Fixed.

roles/telemetry_chargeback/tasks/ingest_loki_data.yml
```diff
@@ -0,0 +1,42 @@
...
+    # Rescue block
+    - name: "Debug failure"
+      ansible.builtin.debug:
+        msg: "{{ loki_response.status | default('N/A') }}"
+
+    # Failure
+    - name: "Report ingestion failure"
+      ansible.builtin.fail:
+        msg: "Ingestion Failed"
+      ignore_errors: false
```
elfiesmelfie, line 42: Combine these two tasks into one.

roles/telemetry_chargeback/tasks/ingest_loki_data.yml
```diff
@@ -0,0 +1,42 @@
+---
+# Ingest data generated by gen_synth_loki_data.yml to Loki
+
+- name: "Ingest data log to Loki via API"
+  block:
+
+    - name: "Read log file content"
+      ansible.builtin.slurp:
+        src: "{{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_data_suffix }}"
+      register: log_file_content
```
elfiesmelfie, line 10: Where is the "{{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_data_suffix }}" file created?

Is there a variable that contains the contents of this file somewhere before it is written to disk?
If so, then you can use that same variable directly instead of reading from the file. It'll reduce the amount of time that it takes to run the test.

roles/telemetry_chargeback/tasks/load_loki_data.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: "Ingest CloudKitty data log for {{ scenario_name }}"
+  ansible.builtin.include_tasks: ingest_loki_data.yml
+
```
elfiesmelfie, line 4: Why is this task on it's own rather than in the run_test_scenarios file?

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -14,6 +14,9 @@
     # Base URL
     loki_base_url: "https://{{ loki_route.stdout }}"
 
+    # Internal Flush URL(Service DNS: https://<service>.<namespace>.svc:3100/flush)
+    ingester_flush_url: "https://cloudkitty-lokistack-ingester-http.{{ cloudkitty_namespace }}.svc:3100/flush"
+
```
elfiesmelfie, line 19: This line should not be added in this PR.
The fact that there's an existing reference to the variable later in this file, implies that there's some PR which adds the ingester_flush_url to the debug, which does not work, since it is not defined at that point.

elfiesmelfie: If the ingester_flush_url is only being used for debugging, don't set it as a separate fact.

If the ingester_flush_url is not being used anywhere else, it should just be removed IMHO

.zuul.yaml
elfiesmelfie, line 1: This is merged to master already, so is not part of this PR

elfiesmelfie (changes requested):

roles/telemetry_chargeback/tasks/load_loki_data.yml
ayefimov-1: ingest_loki_data.yml will be joined by flush_loki_data.yml and retrieve_loki_data.yml in later PRs

PR #380, opened 2026-05-07
Improve RHELOSP-144481 snmpTraps test by adding readiness gates

danpawlik:
The test was failing because it had no synchronization points between enabling snmpTraps, creating the PrometheusRule, disrupting QDR pods, and checking webhook logs. The webhook pod was never receiving alerts from Alertmanager due to timing/race conditions.

Changes:
- Wait for snmp-webhook pod readiness after enabling snmpTraps
- Wait for PrometheusRule to be loaded in Prometheus before disruption
- Verify alert is firing in Alertmanager before checking webhook logs
- Increase QDR disruption window from 30 to 60 iterations
- Increase webhook log check retries from 12 to 30 (5 min)
- Fix bug: rule deletion check was using wrong variable (cmd_output)

Depends-On: https://github.com/infrawatch/feature-verification-tests/pull/379

### DIFF
--- a/roles/test_snmp_traps/tasks/main.yml
+++ b/roles/test_snmp_traps/tasks/main.yml
@@ -35,6 +35,16 @@
       register: cmd_output
       failed_when: cmd_output.rc != 0
 
+    - name: Wait for the snmp-webhook pod to be ready
+      ansible.builtin.shell: >
+        oc wait pod -l app=default-snmp-webhook
+        --for=condition=Ready --timeout=300s
+      register: webhook_wait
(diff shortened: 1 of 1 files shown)


danpawlik:
recheck

elfiesmelfie (changes requested):
Use be Assisted-By, not co-authored-by for Claude

elfiesmelfie (approved):

ayefimov-1 (approved):
LGTM
Change passed in Jenkins

PR #383, opened 2026-05-13
Made changes to the python scripts for consistency and accuracy

ayefimov-1:
- Changed user, project and resource variables consisent with loki/prometheus
- Make python script params more consistent
- Fixed python files to handle mutate: NUMBOOL correctly when calculating rating
- Add scenario name to the db push so that will scenario_name will be filter for db pull

Authored-by @ayefimov-1
Assisted-by: Claude

Closes:
[OSPRH-29232](https://redhat.atlassian.net/browse/OSPRH-29232)

### DIFF
--- a/roles/telemetry_chargeback/files/gen_db_summary.py
+++ b/roles/telemetry_chargeback/files/gen_db_summary.py
@@ -118,11 +118,11 @@ def _apply_mutate(qty: float, mutate: str) -> float:
     elif mutate_upper == "FLOOR":
         return math.floor(qty)
     elif mutate_upper == "NUMBOOL":
-        # If qty equals 0, leave it at 0. Else, set it to 1.
-        return 0.0 if qty == 0 else 1.0
+        # If qty near 0, set it at 0. Else, set it to 1.
+        return 0.0 if abs(qty) < 1e-9 else 1.0
     elif mutate_upper == "NOTNUMBOOL":
-        # If qty equals 0, set it to 1. Else, set it to 0.
-        return 1.0 if qty == 0 else 0.0
+        # If qty near 0, set it to 1. Else, set it to 0.
+        return 1.0 if abs(qty) < 1e-9 else 0.0
     else:  # NONE or any unrecognized value
         return qty
 
@@ -175,8 +175,9 @@ def _parse_numeric(value: Any, default: float = 0) -> float:
 
 def aggregate_rates_by_type(
     pairs: list[tuple[str, str]],
-) -> tuple[dict, float]:
-    sums: defaultdict[str, float] = defaultdict(float)
+) -> tuple[dict, float, dict]:
+    rate_sums: defaultdict[str, float] = defaultdict(float)
+    qty_sums: defaultdict[str, float] = defaultdict(float)
     for _, log_str in pairs:
         try:
             entry = json.loads(log_str)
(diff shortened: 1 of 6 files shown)


ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/files/gen_db_summary.py
```diff
@@ -119,7 +119,7 @@ def _apply_mutate(qty: float, mutate: str) -> float:
         return math.floor(qty)
     elif mutate_upper == "NUMBOOL":
         # If qty equals 0, leave it at 0. Else, set it to 1.
-        return 0.0 if qty == 0 else 1.0
+        return 0.0 if abs(qty) < 1e-9 else 1.0
```
danpawlik, line 122: Please change the comment in L121.
From the other side, should we be so precise here?

roles/telemetry_chargeback/files/gen_db_summary.py
```diff
@@ -237,17 +247,35 @@ def build_summary(pairs: list[tuple[str, str]]) -> dict[str, Any]:
         empty = {"nanosec": None, "begin": None, "end": None}
         time_block = {"begin_step": empty.copy(), "end_step": empty.copy()}
 
-    by_types, total_r = aggregate_rates_by_type(pairs)
+    # Get aggregated data by type
+    by_types, total_r, qty_by_types = aggregate_rates_by_type(pairs)
+
+    # Get overall time range for by_type entries
+    begin_time = first.get("start") if pairs else None
```
danpawlik: nit: IMHO `pairs` needs to exists, otherwise there would be multiple issues in this function.
Maybe better would be: 
```
begin_time = first.get("start")
end_time = last.get("end")
```
if there is no start or end, value would be None.

roles/telemetry_chargeback/files/gen_db_summary.py
```diff
@@ -282,11 +336,21 @@ def main() -> None:
     )
     parser.add_argument(
```
danpawlik, line 340: nit: maybe _str_to_bool would not be needed, if `store_true` we will use.
For example: https://opendev.org/openstack/ci-log-processing/src/branch/master/logscraper/logsender.py#L96

then when `--debug` is used without additional parameter, would be simplier

roles/telemetry_chargeback/files/gen_synth_loki_data.py
```diff
@@ -415,8 +453,10 @@ def main():
     )
     parser.add_argument(
         "--debug",
-        action="store_true",
-        help="Enable debug level logging for verbose output."
+        type=_str_to_bool,
```
danpawlik, line 456: same here with debug

roles/telemetry_chargeback/files/gen_db_summary.py
ayefimov-1: Fixed Comment lines.

While testing I have had NUMBOOL cal fail.  So, I added the above. However, I think there is an error in that  NOTNUMBOOL should have similar "tolerance" value.  Added that.  Thx for the catch.  

Fixed in next push.

roles/telemetry_chargeback/files/gen_db_summary.py
ayefimov-1: Fixed in next push

roles/telemetry_chargeback/files/gen_db_summary.py
ayefimov-1: My goal here was to be able to turn on and off the debug for the entire job with only changing one value in the defaults/main.yml file.  The above will only turn it on or off for the script.

roles/telemetry_chargeback/files/gen_synth_loki_data.py
ayefimov-1: This is as intended , one line change in defaults/main.yml will switch both py scripts to true/false. Seems much easier to have one place to change to get the entire role to debug mode.

danpawlik (approved):

lnatapov (approved):

PR #384, opened 2026-05-14
Fix RHELOSP-176036 and RHELOSP-176038 sensubility health status tests

danpawlik:
Both tests were failing because the script that is running inside the collectd container (collectd_check_health.py; check /etc/collectd-sensubility.conf config file in the collectd container) contains `.*_cron` in SKIP_LIST names that will ignore for checking current status. Instead of fixing the regex `(.*_cron$)`, let's use another container name, that would not affect later tests.


ayefimov-1 (approved):
Looks good!

vyzigold (approved):

danpawlik:
Rebased: ```git pull origin master --rebase``` due there was a "merge" commit

PR #388, opened 2026-05-26
[telemetry_chargeback] Implement test scenario Time-Series Ingestion 

ayefimov-1:
For the DB to work as expected the dataframes must not have overlapping times. This overlapping is currently an issue when generating multiple test scenarios. Implement test scenario Time-Series Ingestion

- Test scenarios will be added to the data base in time sequence
- Each test scenario will not have any time overlap with other scenarios
- Move variables to vars/main.yml
- Readme will be updated in a later PR

Authored-by: @ayefimov-1
AI Assisted by: Claude


ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 openstack_cmd: "openstack"
 
 # Debug mode set
-cloudkitty_debug: false
+cloudkitty_debug: true
```
elfiesmelfie: I prefer to leave debug: false by default.

If you want more information in tests, then configure cloudkitty_debug: true in the playbooks where you're calling this role

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
```diff
@@ -1,15 +1,23 @@
...
   ansible.builtin.set_fact:
-    cloudkitty_data_file: "{{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_data_suffix }}"
-    cloudkitty_synth_totals_file: "{{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_totals_metrics_suffix }}"
-    cloudkitty_test_file: "{{ cloudkitty_scenario_dir }}/{{ scenario_name }}.yml"
+    cloudkitty_data_file: >-
+      {{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_data_suffix }}
+    cloudkitty_synth_totals_file: >-
+      {{ artifacts_dir_zuul }}/{{ scenario_name }}{{ cloudkitty_synth_totals_metrics_suffix }}
+    cloudkitty_test_file: >-
+      {{ cloudkitty_scenario_dir }}/{{ scenario_name }}.yml
```
elfiesmelfie: These should be set in the defaults and then any non-default values should be set in the playbook where the role is imported. In the case of the zuul jobs, this means adding the vars to the ci/vars/ files that we use for osp18

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
```diff
@@ -1,15 +1,23 @@
...
 - name: "Check for preexisting output file"
   ansible.builtin.stat:
     path: "{{ cloudkitty_data_file }}"
   register: file_preexists
 
+- name: "Initialize previous_end_time if not set"
+  ansible.builtin.set_fact:
+    previous_end_time: ""
+  when: previous_end_time is not defined
+
```
elfiesmelfie, line 11: This should be defined as a var on the task.

See https://github.com/infrawatch/feature-verification-tests/blob/master/roles/common/tasks/manifest_tests.yml for an example.

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -30,4 +30,4 @@ lookback: 6
 limit: 50
 
 # List of test scenario files to run
-cloudkitty_test_scenarios: ["test_static.yml" , "test_dyn_basic.yml"]
```
elfiesmelfie, line 33: Are you removing the test_dyn_basic scenario? 
I prefer to keep the original default which is "run all the scenarios", and if our job needs to run fewer, the the playbook in ci/ can set the vars to a reduced set of scenarios.

You need to update the README.md with the new default for this var if you're keeping it.

If you know that the scenario test file is working, you can add it as a separate PR, and update the README and default values for it in a more focussed PR.

roles/telemetry_chargeback/files/gen_synth_loki_data.py
```diff
@@ -127,6 +127,9 @@ def generate_loki_data(
             (newest first, oldest last). If False, sort in ascending order
             (oldest first, newest last). Default is True (descending).
     """
+    # Hardcoded gap between consecutive time steps (in seconds)
+    time_step_gap = 1
```
elfiesmelfie: Add this separately, since it's a minor improvement to the script, and is independant of the time-series ingestion, since it's internal to the script

roles/telemetry_chargeback/files/gen_synth_loki_data.py
```diff
@@ -443,7 +450,7 @@ def main():
         "--descending",
         action="store_true",
         dest="reverse",
-        default=True,
+        default=False,
```
elfiesmelfie: Why does this need to change? You can pass these values into with the ansible task, right?

OR does the end time require a re-work because it only works in one direction?

elfiesmelfie (dismissed):
Please update the PR description to describe the problem that this is solving.

ayefimov-1:
recheck

roles/telemetry_chargeback/defaults/main.yml
ayefimov-1: fixed

roles/telemetry_chargeback/files/gen_synth_loki_data.py
ayefimov-1: This is the time gap between the dataframe time stamps and is required for the this enhancement as I did not want to hard code the time gap.

roles/telemetry_chargeback/files/gen_synth_loki_data.py
ayefimov-1: fixed

roles/telemetry_chargeback/defaults/main.yml
ayefimov-1: fixed

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
ayefimov-1: Fixed. Note that initial value must be "" with second scenario getting value from the results of the first.

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
ayefimov-1: Okay, I understand. Fixed in next pushl

ayefimov-1:
recheck

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
```diff
@@ -18,12 +12,23 @@
       -t "{{ cloudkitty_test_file }}"
       -o "{{ cloudkitty_data_file }}"
       --debug "{{ cloudkitty_debug }}"
+      --end_time "{{ previous_end_time }}"
```
danpawlik, line 20: something that I don't understand, but maybe there si in the code already.
So you set the fact in next task, so the value here is coming from default, and it is always.
Does it mean that the `previous_end_time` is not used anywhere or the next tasks needs to be executed earlier?

ayefimov-1: The dataframes are created in reverse chronological order. I start with the current time and work my way backward for creating synthetic data. 

By default the previous_end_time = "" (see vars/main.yml)
 
If it is empty string, then the script will use time.now as the end_time value and generate the dataframes for the test_scenario dataframe timesteps (chronologically, the last test_scenario)

In the tasks below, the oldest_start_time of the current test scenario is then set to "previous_end_time" for the next test scenario.

vyzigold: @danpawlik These tasks are being run in a loop. If you trace the ansible from here https://github.com/infrawatch/feature-verification-tests/blob/master/roles/telemetry_chargeback/tasks/main.yml#L16 , you'll see. The scripts are kind of generating data and moving backwards in time, but I think the scenarios are set up in a way, where it's hard to calculate beforehand how much data each will generate. Also the data can't overlap, or it screws with the tests. So what this does is, that it generates a chunk of data, then saves the "oldest" datapoint in the next task and that value is then used in the next iteration of the loop, so that the next chunk of data is generated from the "previous_end_time" even more back into history.

It might be a bit more readable, if we could pass the value to these tasks by using vars or maybe set_fact before include_tasks, which would make this all more apparent. But since I think it's not really viable to pre-calculate this "end_time" value, we're stuck with setting it after the "Generate Synthetic Data" task. I don't really have a good suggestion to get around this.

roles/telemetry_chargeback/files/gen_synth_loki_data.py
```diff
@@ -127,6 +127,9 @@ def generate_loki_data(
             (newest first, oldest last). If False, sort in ascending order
             (oldest first, newest last). Default is True (descending).
     """
+    # Hardcoded gap between consecutive time steps (in seconds)
+    time_step_gap = 0
```
vyzigold: So why do we have this variable when it's hardcoded to 0 and never changes?

vyzigold: Also curious, why did it change from 1 to 0?

roles/telemetry_chargeback/tasks/gen_synth_loki_data.yml
danpawlik: It's fine, thanks for explanation.

danpawlik (dismissed):

roles/telemetry_chargeback/files/gen_synth_loki_data.py
ayefimov-1: Good question, this variable is the time gap between dataframe timesteps.  I was not completely sure if the time gap was 0 seconds or 1 second.  So I put it in as a variable. Currently, runs have a variance in the 'rate', and so until I get the correct "rate" results I want to leave it as a var. When 'Rate'  is validating correctly  I will remove this variable.

ayefimov-1:
recheck

lnatapov (dismissed):

lnatapov (dismissed):

roles/telemetry_chargeback/files/gen_synth_loki_data.py
vyzigold: @ayefimov-1

vyzigold (changes requested):
Having a constant of 0, that gets subtracted in the code doesn't make any sense to me. That should either be removed or at least please answer my question before marking the conversation as resolved.

Also are you sure about Emma being on vacation? She surely wasn't yesterday when her review got dismissed, I don't think dismissing reviews should be done lightly.

ayefimov-1:
> Having a constant of 0, that gets subtracted in the code doesn't make any sense to me. That should either be removed or at least please answer my question before marking the conversation as resolved.

    My response  yesterday to your review comment is below, but for some reason  it no longer appears:

  > Good question, this variable is the time gap between dataframe timesteps. I was not completely sure if the time gap was 0 seconds or 1 second. So I put it in as a variable. Currently, runs have a variance in the 'rate', and so until I get the correct "rate" results I want to leave it as a var. When 'Rate' is validating correctly I will remove this variable.

     Removed constant with latest push. Will add it back for debugging purposes if needed

> Also are you sure about Emma being on vacation? She surely wasn't yesterday when her review got dismissed, I don't think dismissing reviews should be done lightly.

  Agreed. dismissing a review should not be done lightly, and  yes, I was wrong about Emma being on vacation! 
  However, all of Emma's review comments/suggestions were addressed.

vyzigold (approved):

vyzigold:
@ayefimov-1 I see the answer in the thread now. Strange that I didn't see it earlier today :+1:

lnatapov (approved):

PR #390, opened 2026-05-28
Reorganize task execution to avoid duplicated RHELOSP-144965 test

danpawlik:
As it was suggested by Emma in the task, we can reorganize the tasks execution to confirm functionality and avoid duplication.


elfiesmelfie (approved):

ayefimov-1 (approved):
Duplicate test removed! Looks good to me.

PR #397, opened 2026-06-04
[OSPRH-30704] Check ceilometer_cpu labels

vyzigold:
Verify that ceilometer_cpu has all the labels we want

### DIFF
--- a/roles/telemetry_verify_metrics/tasks/verify_ceilometer_compute_metrics.yml
+++ b/roles/telemetry_verify_metrics/tasks/verify_ceilometer_compute_metrics.yml
@@ -66,6 +66,27 @@
       changed_when: false
 
     - name: |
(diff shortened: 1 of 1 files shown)


jlarriba:
I agree absolutely with this change and thus I'd approve it, but we have to take in account that the changes that enable project_name and user_name labels on ceilometer_cpu metric merged just today. This means that the fvt will start to fail downstream as soon as we merge this (as they are immediately imported into downstream) until a new ceilometer is built as part of a new compose.

danpawlik (approved):

vyzigold:
@jlarriba Sure, let me put it into draft for now then, so that someone doesn't accidentally merge it.

vyzigold:
A note. The changes to add project_name and user_name were reverted in the telemetry-operator. Plan is to include them again for the first bug fix release of FR6

PR #401, opened 2026-06-22
[telemetry_chargeback] - FVT Run test scenario list chronologically

ayefimov-1:
Test scenario list defined in var "cloudkitty_test_scenarios" is assumed to be in the desired chronological order of testing of scenarios. Since, synthetic data is generated in reverse order, this change makes the job more readable and consistent

### DIFF
--- a/roles/telemetry_chargeback/vars/main.yml
+++ b/roles/telemetry_chargeback/vars/main.yml
@@ -4,6 +4,7 @@ cloudkitty_scenario_dir: "{{ role_path }}/files"
 cloudkitty_synth_script: "{{ role_path }}/files/gen_synth_loki_data.py"
 cloudkitty_data_template: "{{ role_path }}/templates/loki_data_templ.j2"
 cloudkitty_summary_script: "{{ role_path }}/files/gen_db_summary.py"
+cloudkitty_test_file: "{{ cloudkitty_scenario_dir }}/{{ scenario_name }}.yml"
 
 # File naming conventions (internal standardization)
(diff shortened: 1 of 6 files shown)


ayefimov-1:
recheck

ayefimov-1:
recheck

roles/telemetry_chargeback/vars/main.yml
```diff
@@ -11,3 +11,13 @@ cloudkitty_loki_data_suffix: "-loki_data.json"
 cloudkitty_synth_totals_metrics_suffix: "-synth_metrics_summary.yml"
 cloudkitty_loki_totals_metrics_suffix: "-loki_metrics_summary.yml"
 cloudkitty_loki_totals_suffix: "-rating.yml"
+
+cloudkitty_data_file: >-
```
elfiesmelfie: These should be in defaults, not vars, since the artifacts_zuul_dir might not be available. In the case of not being in zuul, I can see the vars being completely overwritten to output the files elsewhere.

elfiesmelfie (commented):
Please move the vars to defaults, I think it makes more sense there, but I won't block based on that.

roles/telemetry_chargeback/vars/main.yml
ayefimov-1: Fixed in  https://github.com/infrawatch/feature-verification-tests/pull/402

danpawlik (approved):

ayefimov-1:
recheck

lnatapov (approved):

PR #402, opened 2026-06-25
[telemetry_chargeback] -FVT Allow artifacts_zuul_dir to be passed to role at run time

ayefimov-1:
Move the output file paths to default dir so that they can be easily overwritten


ayefimov-1:
recheck

elfiesmelfie (approved):

danpawlik (approved):

PR #408, opened 2026-07-01
[telemetry_chargeback] - FVT Compares Loki upload vs download data

ayefimov-1:
Changes include:
- Added loki_diff_db_upload-vs-download.yml task for data validation


roles/telemetry_chargeback/tasks/loki_diff_db_upload-vs-download.yml
```diff
@@ -0,0 +1,44 @@
...
+
+- name: "Check Loki totals file exists"
+  ansible.builtin.stat:
+    path: "{{ cloudkitty_loki_summary_metrics_file }}"
+  register: loki_totals_stat
+
+- name: "TEST Totals files exist {{ scenario_name }}"
+  ansible.builtin.assert:
+    that:
+      - synth_totals_stat.stat.exists | default(false)
```
danpawlik: remove the defaults here. The stat tasks in L3 and L7 does not have condition, so those vars would always be present.

roles/telemetry_chargeback/files/gen_db_summary.py
```diff
@@ -277,9 +282,7 @@ def build_summary(pairs: list[tuple[str, str]]) -> dict[str, Any]:
             "log_count": log_count,
             "total_rating": round(total_r, 4),
         },
-        "by_type": {
-            "rate": rate_list,
-        },
+        "rate_by_type": rate_list,
```
danpawlik, line 285: just curious: change for the the type is ok or it is a mistake?
asking, because: `by_type['rate']` would be now `rate_by_type`.

ayefimov-1: that fine.  This is output for the summary and  matches the openstack rate command output  more closely

roles/telemetry_chargeback/tasks/loki_diff_db_upload-vs-download.yml
ayefimov-1: fixed in next push.

danpawlik (approved):

lnatapov (approved):

PR #411, opened 2026-07-10
Delete obsolete jobs

vyzigold:
Remove jobs that we no longer use and which we don't plan to start using again.

functional-tests-on-osp18 was just an alias to functional-tests-osp18 likely introduced due to a mistake. It's no longer needed.

functional-autoscaling-tests-osp18, functional-graphing-tests­osp18 and functional-metric-verification-tests-osp18 were all merged into the functional-tests-osp18 and so aren't needed anymore.

None of the jobs were ever executed on the current Zuul tenant, meaning they weren't run in a long time

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -126,31 +126,6 @@
       - roles/test_verify_email/.*
 
 - job:
-    name: functional-tests-on-osp18
-    parent: functional-tests-osp18
-    extra-vars: *functional_autoscaling_extra_vars
-    description: |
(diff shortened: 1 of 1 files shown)


vyzigold:
Also, looking at the jobs in https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/jobs , all the deleted jobs show as leafs in the graph, so it's not like anything is inheriting from them.

danpawlik (approved):

elfiesmelfie (approved):

PR #415, opened 2026-07-14
[telemetry_chargeback] [FVT] Add Loki ring convergence check and update defaults

ayefimov-1:
Fix for Loki returning empty query results caused by ring partitioning 
- Wait for the Loki gossip ring to converge before ingesting synthetic data
- Preventing empty query results caused by ring partitioning. 
- remove lookback, increase limit to 1500

Authored-by: @ayefimov-1
AI Assisted by: Claude

### DIFF
--- a/roles/telemetry_chargeback/defaults/main.yml
+++ b/roles/telemetry_chargeback/defaults/main.yml
@@ -26,11 +26,10 @@ cloudkitty_namespace: "openstack"
 openstackpod: "openstackclient"
 
 # Time window settings
-lookback: 6
-limit: 50
+limit: 1500
 
 # List of test scenario files to run
-cloudkitty_test_scenarios: ["test_static" , "test_dyn_basic"]
+cloudkitty_test_scenarios: ["test_static", "test_dyn_basic"]
 
 # Output file paths
 # ALL variables using {{ artifacts_dir_zuul }} should be defined in defaults/main.yml
--- a/roles/telemetry_chargeback/tasks/setup_loki_env.yml
+++ b/roles/telemetry_chargeback/tasks/setup_loki_env.yml
@@ -57,3 +57,25 @@
     cmd: |
       oc extract {{ ca_configmap }} --to={{ local_cert_dir }} --confirm -n {{ cloudkitty_namespace }}
   changed_when: true
+
+# Wait for Loki gossip ring to converge before ingesting data.
+# When the ring hasn't converged, ingesters accept writes but are
+# invisible to queriers, causing queries to return empty results.
+- name: "Wait for Loki ring to converge"
+  ansible.builtin.uri:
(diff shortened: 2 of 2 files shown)


ayefimov-1:
recheck

roles/telemetry_chargeback/defaults/main.yml
```diff
@@ -26,11 +26,10 @@ cloudkitty_namespace: "openstack"
 openstackpod: "openstackclient"
 
 # Time window settings
-lookback: 6
-limit: 50
+limit: 1500
```
danpawlik, line 29: I don't see lookback is used anywhere, so that is correct.
About `limit`, IMHO - README file should be also updated.

danpawlik (approved):
README can be updated later

roles/telemetry_chargeback/tasks/setup_loki_env.yml
```diff
@@ -57,3 +57,37 @@
...
+# from inside the cluster.
+- name: "Create cert directory in OpenStack Client pod"
+  ansible.builtin.command:
+    cmd: "oc exec -n {{ cloudkitty_namespace }} {{ openstackpod }} -- mkdir -p {{ remote_cert_dir }}"
+  changed_when: false
+
+- name: "Copy certificates to OpenStack Client pod"
+  ansible.builtin.command:
+    cmd: "oc cp {{ local_cert_dir }}/. {{ cloudkitty_namespace }}/{{ openstackpod }}:{{ remote_cert_dir }}/"
+  changed_when: false
```
vyzigold, line 74: There is the same set of tasks in https://github.com/infrawatch/feature-verification-tests/blob/master/roles/telemetry_chargeback/tasks/flush_loki_data.yml#L6C1-L16C25 . This may be an opportunity for some future refactoring. (pulling them out into a separate task file or ensuring these are executed once before anything else that needs them). There are other tasks accessing the Loki API in different ways like https://github.com/infrawatch/feature-verification-tests/blob/master/roles/telemetry_chargeback/tasks/retrieve_loki_data.yml#L22C1-L42C16 , which we may be able to modify to be similar and use the same certs etc....

Just a food for thought though. I won't block on this now.

vyzigold (approved):

roles/telemetry_chargeback/tasks/setup_loki_env.yml
ayefimov-1: https://redhat.atlassian.net/browse/OSPRH-29987 created

roles/telemetry_chargeback/defaults/main.yml
ayefimov-1: README File changes in https://github.com/infrawatch/feature-verification-tests/pull/421

PR #420, opened 2026-07-27
[telemetry_chargeback] Price needs to be calculated for each synthetic dataframe

ayefimov-1:
gen_synth_loki_data.py:
- Compute price at generation time: price = unit_cost * qty_mutated.
- Add _apply_mutate() for NUMBOOL/CEIL/FLOOR transforms.
- Remove 2 hour offset

gen_db_summary.py:
- Removed _apply_mutate() function — mutate transformations (CEIL, FLOOR, NUMBOOL, NOTNUMBOOL)
- Simplified aggregate_rates_by_type() — it now rate_sums[mtype] += price
- Added unit_cost to REQUIRED_KEYS — entries now must include unit_cost alongside price.

Update test scenarios: 
- 1-hour window per scenario
- 300s steps
- realistic per unit costs
- Remove factor/offset from all files (currently not used in rating)
- Remove "price" from test scenario definition as gen_synth script wil generate price

Authored-by: @ayefimov-1
AI Assisted by: Claude

### DIFF
--- a/roles/telemetry_chargeback/defaults/main.yml
+++ b/roles/telemetry_chargeback/defaults/main.yml
@@ -26,11 +26,10 @@ cloudkitty_namespace: "openstack"
 openstackpod: "openstackclient"
 
 # Time window settings
-lookback: 6
-limit: 50
+limit: 1500
 
 # List of test scenario files to run
-cloudkitty_test_scenarios: ["test_static" , "test_dyn_basic"]
+cloudkitty_test_scenarios: ["test_static", "test_dyn_basic"]
 
 # Output file paths
 # ALL variables using {{ artifacts_dir_zuul }} should be defined in defaults/main.yml
--- a/roles/telemetry_chargeback/files/gen_db_summary.py
+++ b/roles/telemetry_chargeback/files/gen_db_summary.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """
 Parse Loki JSON (or text) into [timestep, log_entry] pairs, then emit a YAML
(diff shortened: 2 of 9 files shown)


ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

ayefimov-1:
recheck

jlarriba:
I think it is still calculating the rate wrong. The diff test (loki_diff_db_upload-vs-download.yml) still passes because both the upload and download paths run through gen_db_summary.py and compute the same wrong value. But the total_rate_scenario in the summary YAML will be incorrect.

You should either have gen_db_summary.py use unit_cost instead of price in the rate formula or use price directly as the rate without multiplying by qty_mutated again. 

That will give you the correct rate.

ayefimov-1:
> I think it is still calculating the rate wrong. The diff test (loki_diff_db_upload-vs-download.yml) still passes because both the upload and download paths run through gen_db_summary.py and compute the same wrong value. But the total_rate_scenario in the summary YAML will be incorrect.
> 
> You should either have gen_db_summary.py use unit_cost instead of price in the rate formula or use price directly as the rate without multiplying by qty_mutated again.
> 
> That will give you the correct rate.

Fixed gen_db_summary.py

jlarriba (approved):

lnatapov (approved):

PR #425, opened 2026-08-06
Adjust test tasks names so they will report to Polarion

lnatapov:
Adjust test tasks names so they will report to Polarion


elfiesmelfie (approved):

vyzigold (approved):
I have no problem with this, but I don't see what's the difference. Why did the previous version didn't report to polarion and this does?

PR #426, opened 2026-08-11
Avoid blocking QDR interrupt loops on oc delete --wait

danpawlik:
Use --wait=false so interconnect pod churn does not stall for minutes per iteration under load, which was hanging stf_functional_tests.

Assisted-by: Cursor <cursoragent@cursor.com>


elfiesmelfie (approved):
That makes sense.

We just want to make sure the pod stays down for long enough to detect an interrupt. We don't need to wait for it to be gone to move on with the test.

ayefimov-1 (approved):

